PR #7, opened 2016-03-30
Add property "dn" and APIs for resource status

amittbose:


aim/api/status.py
```diff
@@ -0,0 +1,101 @@
...
+    def is_error(self):
+        """Returns True if current status is an error state."""
+        raise
+
+
+class AciStatus(Status):
+    """Status of an AIM resource that is mapped to ACI object.
+
+    Following attributes are available:
+    * create_status - Indicates whether ACI object was created or not
```
rkukura: Is this intended to reflect whether the object has been created in the APIC? Does it also reflect whether the APIC has been updated with any changes made to the object after it was created? I was expecting something called sync_status that reflected both - basically indicating whether all pending changes to the resource have been synced to the APIC. Would it make sense to rename/redefine this to sync_status?

aim/aim_manager.py
```diff
@@ -150,6 +151,21 @@ def find(self, context, resource_class, **kwargs):
...
 
+    def get_status(self, context, resource):
+        """Get status of an AIM resource, if any."""
+
+        if isinstance(resource, api_res.AciResourceBase):
+            # TODO(amitbose) Fetch status from DB
+            return api_status.AciStatus()
+        return None
+
+    def update_status(self, context, resource, status):
```
rkukura: I am I correct that this is for use by the AID, not be northbound API users?

aim/aim_manager.py
```diff
@@ -150,6 +151,21 @@ def find(self, context, resource_class, **kwargs):
                 self._make_resource(context.db_session, resource_class, obj))
         return result
 
+    def get_status(self, context, resource):
```
rkukura: I'm curious why this is a separate AimManager call that has to be made after the Resource is retrieved, and not available as an attribute of Resource?

amittbose: Status is operational state while the resource represents the configuration/desired state. I think it makes sense to keep the two separate. Note, that, to get the status of a resource, you don't need to "retrieve" the resource from AimManager by calling get(). You can just create the Python object for a resource with the minimum required attributes and pass it to get_status().

aim/aim_manager.py
amittbose: That is correct.

aim/api/status.py
amittbose: I was discussing this with Sumit earlier today, and it seemed that setting the status of a resource to "BUILD" when changes are made to it was probably not useful. Also this situation is likely to be very transient. But I guess if that sync fails for some reason (e.g. update violates some check in ACI), then it would make sense to report that. I'll update the diffs.

amittbose:
Addressed Bob's comments.

aim/aim_manager.py
rkukura: OK, thanks. Unless things change, my code will always be getting the DN and the status at the same time, but I understand the decision to separate these.

aim/api/resource.py
```diff
@@ -66,6 +85,8 @@ class BridgeDomain(ResourceBase):
     # Attrbutes completely managed by the DB (eg. timestamps)
     db_attributes = []
 
+    _aci_mo_name = 'fvBD'
```
ivar-lazzaro: One MO name is not always enough (In fact, it is not even for a simple BD). In our case, the Bridge Domain is composed of fvBD and fvRsCtx on APIC which means that two DNs should be extracted from it (and shown by Neutron).

Should we call the converter in order to retrieve the backend identification? This would also have the extra advantage of removing ACI specific information from the AIM library (the converter can eventually be a pluggable component).

amittbose: This doesn't make any sense to me. A single AimResource maps to a single ACI object, and therefore a single DN. BridgeDomain maps to fvBD, that's it. fvRsCtx is a child "object" that hangs off fvBD and doesn't identify the ACI object that BridgeDomain corresponds to. I wouldn't treat relationship MOs like fvRsCtx the same as MOs like fvBD which represent independent, concrete concepts. fvRsCtx is derived information that you can retrieve from the fvBD.
As far as the end user (in Neutron) is concerned, there is no value to showing the DN of the fvRsCtx object.

ivar-lazzaro: Neutron shows DNs for the only purpose of debugability, giving back one part of the story is not very helpful IMHO. An AIM BridgeDomain can't be reconstructed by simply retrieving it's parent object.
But to your point, we can make it a special case of "hidden detail" for all the RS objects, but not all the type of children (like fvSubnet, for instance).

This would be acceptable as long as the RSs' DN can be predicted from the parents' DN. Which is possible with the fvRsCtx case, not sure for rest of the model though.

Lastly, I still believe that using the converter to retrieve the DN (and removing _aci_mo_name) is a good option for separation of concerns, which limits APICAPI presence in the least number of modules (all in the aci_universe package). But this is by no means urgent (and should be fairly easy to do later).

amittbose: > An AIM BridgeDomain can't be reconstructed by simply retrieving it's parent object.
That was NEVER the intention of the DN property. It only serves to provide a pointer to the mapped ACI object, thus enabling the consumer to explore any additional information they are interested in by interacting directly with APIC.

> "hidden detail" for all the RS objects
Why? One of the reasons we have AIM is to hide these very ugly details from the AIM consumer. I would like to be conservative and avoid unnecessary details through the AIM APIs.

> converter to retrieve the DN  .. which limits APICAPI presence in the least number of modules (all in the aci_universe package).
Actually, I think it should be the other way around - the converter should get the DN from AIM resource instead of creating one of its own. DNs for child objects should be derived from that DN. Frankly, I'm not too concerned with apicapi presence in too many modules; in any case my intention is to tease the MO definitions out of apicapi in to AIM. So this dependency will eventually go away.

ivar-lazzaro: > That was NEVER the intention of the DN property

I'm not saying we intended scripts or even humans to do that. I'm saying that giving the whole side-effects picture is better for debugability, which is the intention of the DN(s) attribute exposed by Neutron (as far as I recall from our discussions, which might be wrong/changed).

> Why? One of the reasons we have AIM is to hide these very ugly details from the AIM consumer

We do hide them. The DN(s) attribute is not for the northbound project to understand.

> Actually, I think it should be the other way around

I was under the impression that we wanted to separate ACI specifics from the AIM library and delegate them to the AID module (when we discussed moving "uglyness" in the agent). If that's not the case anymore, then it's a better solution to decorate the AIM resources.

amittbose: Honestly speaking we can't paint the whole picture for debugging anyways because there are so many things in ACI that are beyond AIM control. Instead we should strive to provide concise meaningful information that will *enable* the end user to debug ACI.

Regarding ugliness: in the case of fvBD, fvRsCtx is actually the ugly part. So it should not be part of the AIM resource because we don't want AIM clients to be bothered about it. fvRsCtx exists only in AID right now, and that is great. Exposing the DN of fvRsCtx through AIM will breach the abstraction that AIM is trying to provide, no?

ivar-lazzaro: > Honestly speaking we can't paint the whole picture for debugging anyways

Showing the user the whole picture of what *we* control (fvRsCtx in this case) would be possible. But I understand that you want to limit the number of DNs from a single AIM resource to just one. As long as "hiding" the RSs will suffice to reach this goal I'm ok with it. Especially if all the RS DNs are just like fvRsCtx, predictable from the parent.

> Regarding ugliness: in the case of fvBD, fvRsCtx is actually the ugly part.

The ugly part is every ACI specific information IMHO. By confining it to the agent we can eventually enable different backends (or at least that was the claim, at least as long as it doesn't slow down our main goal).

> Exposing the DN of fvRsCtx through AIM will breach the abstraction that AIM is trying to provide, no?

As I said above, I'm OK hiding the RS DNs from Neutron. Even so, I think the answer to this question is no. We break the abstraction if we force the AIM user (Neutron in this case) to understand the ACI specifics, which we are not since Neutron doesn't have to understand DNs (any DN in fact) but just dumbly show them to the human user.

I see the DN as a debug utility for humans/test scripts. Not something that undermines the abstraction. Hypothetically, a backend other than ACI could show something different in the DN field with a similar meaning.

amittbose: I don't see any value in exposing the fvRsCtx DN for debugging. It is a detail that the AIM user should not be bothered with.

ivar-lazzaro: In that case, let's merge this as is. Let's take a decision on whether we want to move the DN calculation in the converter for concern separation though, and revisit this once we do.

PR #10, opened 2016-04-05
Agent end-to-end integration

ivar-lazzaro:
Put together all the pieces for end to end AID integration

### DIFF
--- a/aim/api/resource.py
+++ b/aim/api/resource.py
@@ -47,8 +47,9 @@ class ResourceBase(object):
                       if kwargs.get(k) is None and k not in defaults]
         if unset_attr:
             raise exc.IdentityAttributesMissing(attr=unset_attr)
-        for k, v in defaults.iteritems():
-            setattr(self, k, v)
+        if kwargs.pop('set_default', True):
+            for k, v in defaults.iteritems():
+                setattr(self, k, v)
         for k, v in kwargs.iteritems():
             setattr(self, k, v)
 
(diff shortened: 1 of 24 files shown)


ivar-lazzaro:
For now, I have pushed a proposal patch that leaves the TreeMapping entry for empty tenants so that AID can take care of the cleanup properly.
This approach will guarantee that the Tenant gets eventually cleaned up even if all the AIDs were down at the time of the deletion.

However, leaving those stale entries cause more issues than just some wasted disk... One being that the Agents will still monitor that tenant, even at the cost of *deleting if someone creates it manually*. This is something we need to solve in the final solution.

aim/api/resource.py
```diff
@@ -47,8 +47,9 @@ def __init__(self, defaults, **kwargs):
                       if kwargs.get(k) is None and k not in defaults]
         if unset_attr:
             raise exc.IdentityAttributesMissing(attr=unset_attr)
-        for k, v in defaults.iteritems():
-            setattr(self, k, v)
+        if kwargs.pop('set_default', True):
```
amittbose: can we prefix meta-attributes like these with '_' ?

aim/db/hashtree_db_listener.py
```diff
@@ -50,25 +51,20 @@ class DummyContext(object):
...
 
         # Finally save the modified trees
         if upd_trees:
             self.tt_mgr.update_bulk(ctx, upd_trees)
-        if del_trees:
-            self.tt_mgr.delete_bulk(ctx, del_trees)
+
+    def _get_resource_tenant(self, resource):
+        return (resource.name if isinstance(resource, api_res.Tenant) else
+                getattr(resource, 'tenant_name', None))
```
amittbose: I prefer to avoid assuming that the tenant attribute will have a specific name. Why not just change AimHashTreeMaker.get_tenant_key() to return the tenant-name instead?

aim/db/tree_model.py
```diff
@@ -211,6 +232,30 @@ def get_tenant_key(self, resource):
...
+
+    @staticmethod
+    def tenant_key_funct(key):
+        """Key funct for Tree Maker
+
+        Utility function for TreeManager initialization
+        :param tree:
+        :return:
+        """
+        return api_res.__name__ + '.Tenant|' + key,
```
amittbose: Prefer that you factored out lines 199-200 and use that here

aim/api/resource.py
ivar-lazzaro: Yeah agreed

aim/db/tree_model.py
ivar-lazzaro: ok

aim/db/hashtree_db_listener.py
ivar-lazzaro: ok

PR #11, opened 2016-04-06
Define resources for subnet, VRF, endpoint-groups

amittbose:
DB-model for endpoint-groups is missing contracts.
Changes to ACI-converter are also missing.

### DIFF
--- a/aim/api/resource.py
+++ b/aim/api/resource.py
@@ -39,6 +39,9 @@ class ResourceBase(object):
     Class property 'db_attributes' gives a list of resource attributes
     that are managed by the database layer, eg: timestamp, incremental counter.
     """
+
+    db_attributes = []
+
     def __init__(self, defaults, **kwargs):
         unset_attr = [k for k in self.identity_attributes
                       if kwargs.get(k) is None and k not in defaults]
@@ -86,7 +89,6 @@ class Tenant(AciResourceBase):
 
     identity_attributes = ['name']
     other_attributes = ['display_name']
-    db_attributes = []
 
     _aci_mo_name = 'fvTenant'
     _tree_parent = None
@@ -109,8 +111,6 @@ class BridgeDomain(AciResourceBase):
                         'limit_ip_learn_to_subnets',
                         'l2_unknown_unicast_mode',
                         'ep_move_detect_mode']
-    # Attrbutes completely managed by the DB (eg. timestamps)
-    db_attributes = []
 
     _aci_mo_name = 'fvBD'
     _tree_parent = Tenant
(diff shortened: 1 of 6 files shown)


aim/api/resource.py
```diff
@@ -148,3 +148,86 @@ def is_down(self):
...
+
+    identity_attributes = ['tenant_name', 'bd_name', 'gw_ip_mask']
+    other_attributes = ['scope',
+                        'display_name']
+
+    _aci_mo_name = 'fvSubnet'
+    _tree_parent = BridgeDomain
+
+    def __init__(self, **kwargs):
+        super(Subnet, self).__init__({}, **kwargs)
```
ivar-lazzaro: Just for reference, every time we add a AIM resource we need to specify meaningful default values for all the attributes that will show up in APIC. This will help consistent conversion (also, every time we do a GET on a AIM object, DB default values will show up... Which might not be meaningful at all).

aim/db/model_base.py
```diff
@@ -47,6 +50,14 @@ class HasId(object):
                    default=utils.generate_uuid)
 
 
+class HasDisplayName(object):
+    display_name = sa.Column(sa.String(256))
+
+
+class HasTenantNameKey(object):
+    tenant_name = name_column(primary_key=True)
```
ivar-lazzaro: Should this be a foreign key? I thought we weren't going to support "partial trees" in AIM (as in, objects pointing missing ones).

aim/db/models.py
```diff
@@ -18,16 +18,67 @@
...
     ep_move_detect_mode = sa.Column(sa.String(16))
+
+
+class Subnet(model_base.Base, model_base.HasTenantNameKey,
+             model_base.AttributeMixin):
+    """DB model for Subnet."""
+
+    __tablename__ = 'aim_subnets'
+
+    bd_name = model_base.name_column(primary_key=True)
```
ivar-lazzaro: Same consideration as above about FKs, should (tenant_name, bd_name) point to an actual BD?

aim/api/resource.py
```diff
@@ -86,7 +89,6 @@ class Tenant(AciResourceBase):
 
     identity_attributes = ['name']
     other_attributes = ['display_name']
-    db_attributes = []
 
     _aci_mo_name = 'fvTenant'
     _tree_parent = None
```
rkukura: Should the Neutron (and eventually GBP) ensure_tenant() stuff be creating Tenant resources if they don't already exist?

aim/api/resource.py
```diff
@@ -148,3 +148,86 @@ def is_down(self):
...
+
+    def __init__(self, **kwargs):
+        super(Subnet, self).__init__({}, **kwargs)
+
+    @staticmethod
+    def to_gw_ip_mask(gateway_ip_address, prefix_len):
+        return '%s/%d' % (gateway_ip_address, prefix_len)
+
+
+class VRF(AciResourceBase):
```
rkukura: Is "VRF" the proper ACI term, or is "Context"?

aim/api/resource.py
amittbose: Northbound clients are not required to create the Tenant resource, but it wouldn't harm anything if you created one. I'll leave up to you to make a choice.

aim/api/resource.py
amittbose: "Context" is what I see used internally in ACI (e.g visore) and in the REST APIs. APIC GUI and documentation refer to it as "VRF". I picked VRF because it was the publicly documented term. "Context" is too generic.

aim/db/model_base.py
amittbose: OK.

amittbose:
Review comments addressed.

aim/api/resource.py
ivar-lazzaro: If Tenant name becomes a foreign Key then a Tenant resource will be mandatory, right? Will it be created implicitly?

aim/api/resource.py
ivar-lazzaro: +1

aim/api/resource.py
ivar-lazzaro: I see now that first level objects don't have FK constraints... This could be an issue for the synchronization, but let's merge it for now and see if we can circle this problem around without having the Northbound project creating a tenant.

PR #12, opened 2016-04-08
Add DB model for EPG contracts

amittbose:
Have used an auto-increment column on the EPG table
to generate an internal ID that will be referenced
by the contracts table. This is needed for resources
that have non-scalar attributes (e.g. list of contracts).
Having an internal ID avoids having primary
keys that use many columns in the table that contains
the non-scalar values. Have updated other ACI model
classes with this internal-ID.

### DIFF
--- a/aim/db/models.py
+++ b/aim/db/models.py
@@ -14,23 +14,38 @@
 #    under the License.
 
 import sqlalchemy as sa
+from sqlalchemy import orm
 
 from aim.db import model_base
 
 
-class Tenant(model_base.Base, model_base.HasName,
-             model_base.HasDisplayName, model_base.AttributeMixin):
+def to_tuple(obj):
+    return obj if isinstance(obj, tuple) else (obj,)
+
+
+def uniq_column(table, *args):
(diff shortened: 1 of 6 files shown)


aim/db/models.py
```diff
@@ -40,57 +51,104 @@ class BridgeDomain(model_base.Base, model_base.HasName,
...
+        ) + to_tuple(model_base.Base.__table_args__)
+
+    app_profile_name = model_base.name_column(nullable=False)
 
-    app_profile_name = model_base.name_column(primary_key=True)
     bd_name = model_base.name_column()
     bd_tenant_name = model_base.name_column()
-    # TODO(amitbose) Map contract names
+
+    contract = orm.relationship(EndpointGroupContract,
```
ivar-lazzaro: this should be contracts*

amittbose:
Addressed review comments:
* Renamed 'contract' to 'contracts' in EPG DB model
* Added indices for columns that have unique-constraint

For now I'm keeping the surrogate 'id' column as the primary key because I convinced myself by reading online discussions that doing so is a better approach :)

PR #23, opened 2016-07-07
Add resource and DB model for contracts/filters

amittbose:

### DIFF
--- a/aim/api/resource.py
+++ b/aim/api/resource.py
@@ -256,3 +256,113 @@ class EndpointGroup(AciResourceBase):
                                              'provided_contract_names': [],
                                              'consumed_contract_names': []},
                                             **kwargs)
+
+
+class Filter(AciResourceBase):
+    """Resource representing a contract filter in ACI.
+
+    Identity attributes: name of ACI tenant and name of filter.
+    """
+
+    identity_attributes = ['tenant_name', 'name']
+    other_attributes = ['display_name']
+
+    _aci_mo_name = 'vzFilter'
+    _tree_parent = Tenant
+
+    def __init__(self, **kwargs):
+        super(Filter, self).__init__({}, **kwargs)
+
+
+class FilterEntry(AciResourceBase):
+    """Resource representing a classifier entry of a filter in ACI.
+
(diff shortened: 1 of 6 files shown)


aim/api/resource.py
```diff
@@ -256,3 +256,113 @@ def __init__(self, **kwargs):
...
+
+
+class FilterEntry(AciResourceBase):
+    """Resource representing a classifier entry of a filter in ACI.
+
+    Identity attributes: name of ACI tenant, name of filter and name of entry.
+
+    Values for classification fields may be integers as per standards
+    (e.g. ip_protocol = 6 for TCP, 17 for UDP), or special strings listed
+    below. UNSPECIFIED may be used to indicate that a particular
```
ivar-lazzaro: Is a fully "unspecified" filter working as a "allow all"?

amittbose: That's right - an entry like FilterEntry(tenant_name=..., filter_name=..., name=...) is an allow-all entry.

aim/db/models.py
```diff
@@ -162,3 +162,120 @@ def to_attr(self, session):
...
+
+
+class Filter(model_base.Base, model_base.HasAimId,
+             model_base.HasName, model_base.HasDisplayName,
+             model_base.HasTenantName, model_base.AttributeMixin):
+    """DB model for Filter."""
+
+    __tablename__ = 'aim_filters'
+    __table_args__ = (uniq_column(__tablename__, 'tenant_name', 'name') +
+                      to_tuple(model_base.Base.__table_args__))
```
ivar-lazzaro: would a Entry backref be useful here?

amittbose: I don't see the need yet. This is a simple parent-child relationship in the tree, and so far there hasn't been a need to create a backref to the parent from the child.

aim/api/resource.py
ivar-lazzaro: sounds good

aim/db/models.py
ivar-lazzaro: yeah I guess only faults and status have such a thing, no need to add a backref if we don't believe there's a use for that.

PR #24, opened 2016-07-12
DB based configuration for APIC/AIM options

ivar-lazzaro:


ivar-lazzaro:
this needs more work for reactive conf

PR #25, opened 2016-07-18
Add converter rules for contracts

amittbose:
For the most part these rules are pretty straight-forward
except for handling of in/out filters in subjects. The
AIM-to-ACI converter translates these to pseudo ACI MOs like
vzRsFiltAtt__In/vzRsFiltAtt__Out, but the hash-tree nodes
have the correct ACI MOs (vzSubj -> vzInTerm -> vzRsFiltAtt,
vzSubj -> vzOutTerm -> vzRsFiltAtt).
A few small changes were needed to ACI universe code to
handle the use-cases presented by vzInTerm/vzOutTerm.

### DIFF
--- a/aim/tools/cli/debug/hashtree.py
+++ b/aim/tools/cli/debug/hashtree.py
@@ -1,0 +1,49 @@
+# Copyright (c) 2016 Cisco Systems
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


aim/tools/cli/debug/hashtree.py
```diff
@@ -0,0 +1,49 @@
...
+from aim.common.hashtree import exceptions as h_exc
+from aim import context
+from aim.db import api
+from aim.db import tree_model
+from aim.tools.cli.groups import aimcli
+
+
+@aimcli.aim.group(name='hashtree')
+@click.pass_context
+def hashtree(ctx):
```
ivar-lazzaro: don't see the CLI commands covered by UTs, am I missing them?

amittbose: Did realize that debug CLI commands also had UTs. Will add it in a follow-up change if that is ok with you.

PR #26, opened 2016-07-25
AIM resource for endpoints

amittbose:
Adding a simple resource object for endpoints
which mainly stores endpoint association to EPG.

Also added a utility function from_dn() to
construct an AIM resource from a DN string.

### DIFF
--- a/aim/db/models.py
+++ b/aim/db/models.py
@@ -280,3 +280,23 @@ class ContractSubject(model_base.Base, model_base.HasAimId,
                 attr = 'bi_filters'
             res_attr.setdefault(attr, []).append(f.name)
         return res_attr
+
+
+class Endpoint(model_base.Base, model_base.HasDisplayName,
+               model_base.AttributeMixin):
+    """DB model for Endpoint."""
+
+    __tablename__ = 'aim_endpoints'
+    __table_args__ = (
+        (sa.ForeignKeyConstraint(
+            ['epg_tenant_name', 'epg_app_profile_name', 'epg_name'],
+            ['aim_endpoint_groups.tenant_name',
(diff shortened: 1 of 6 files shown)


aim/db/models.py
```diff
@@ -280,3 +280,23 @@ def to_attr(self, session):
...
+        (sa.ForeignKeyConstraint(
+            ['epg_tenant_name', 'epg_app_profile_name', 'epg_name'],
+            ['aim_endpoint_groups.tenant_name',
+             'aim_endpoint_groups.app_profile_name',
+             'aim_endpoint_groups.name'],
+            name='fk_epg'),) +
+        to_tuple(model_base.Base.__table_args__))
+
+    uuid = sa.Column(sa.String(36), primary_key=True)
+    epg_tenant_name = model_base.name_column()
```
ivar-lazzaro: why not using tenant_name as in other tables?

amittbose: Because this is the EPG's tenant.

ivar-lazzaro: Could this be different than the EP tenant? Are you thinking of shared networks case?

amittbose: At this point, I don't want to make any assumption about EP's tenant w.r.t. EPG's tenant. This field is just telling what tenant the associated EPG belongs to, nothing more. The use of 'tenant_name' in other objects is different because it basically identifies the ACI tenant of the corresponding ACI object. That doesn't apply to endpoints in ACI. So I think it is more appropriate to name the field epg_tenant_name.

ivar-lazzaro: Makes sense, thanks for the clarification

PR #28, opened 2016-08-02
Fixes for faults reported for filter relations

amittbose:
Faults reported for subject-to-filter relations
were either ignored completely or raising exceptions
in AID. This change fixes fault reporting for
such cases where the fault parent in ambiguous by
using newly added utility methods in apicapi
DNManager.


amittbose:
Added UTs.

PR #37, opened 2016-09-12
Monitored universe

ivar-lazzaro:


aim/agent/aid/universes/aci/converter.py
```diff
@@ -351,13 +349,11 @@ def fv_rs_dom_att_converter(object_dict, otype, helper,
...
                   'converter': fvRsProv_converter},
                  {'resource': resource.ExternalNetwork,
-                  'converter': fvRsProv_Ext_converter,
-                  'convert_pre_existing': True}],
+                  'converter': fvRsProv_Ext_converter}],
     'fvRsCons': [{'resource': resource.EndpointGroup,
                   'converter': fvRsCons_converter},
                  {'resource': resource.ExternalNetwork,
-                  'converter': fvRsCons_Ext_converter,
-                  'convert_pre_existing': True}],
```
amittbose: I think we still need to make a distinction between relations that we want to configure and others that we want to monitor

aim/db/hashtree_db_listener.py
```diff
@@ -36,44 +36,64 @@ def on_commit(self, session, added, updated, deleted):
...
-                    (2 if isinstance(res, aim_status.OperationalResource)
-                     else 0)].append(res)
+                updates_by_tenant.setdefault(
+                    key, {conf: ([], []), monitor: ([], []), oper: ([], [])})
+                if isinstance(res, aim_status.OperationalResource):
+                    # Operational Tree
+                    updates_by_tenant[key][oper][tree_index].append(res)
+                elif getattr(res, 'monitored', None):
+                    # Monitored Tree
+                    updates_by_tenant[key][monitor][tree_index].append(res)
```
amittbose: So monitored AIM resources always go to the MONITORED tree (and are therefore not synced to ACI). But an AIM resource can convert to several relation objects as well. If these are put in the MONITORED tree, then the relations won't get synced?

aim/db/hashtree_db_listener.py
```diff
@@ -36,44 +36,64 @@ def on_commit(self, session, added, updated, deleted):
...
             try:
-                ttree = self.tt_mgr.get(ctx, tenant)
-                ttree_operational = self.tt_mgr.get(ctx, tenant,
-                                                    operational=True)
+                ttree = self.tt_mgr.get(ctx, tenant, tree=conf)
+                ttree_operational = self.tt_mgr.get(ctx, tenant, tree=oper)
+                ttree_monitor = self.tt_mgr.get(ctx, tenant, tree=monitor)
             except hexc.HashTreeNotFound:
                 ttree = htree.StructuredHashTree()
                 ttree_operational = htree.StructuredHashTree()
```
amittbose: Are you sure you want to reset all the 3 trees even if only of them is not present?

aim/db/migration/alembic_migrations/versions/8e313fbeb93b_l3out.py
```diff
@@ -50,6 +50,7 @@ def upgrade():
         sa.Column('pre_existing', sa.Boolean),
```
amittbose: remove 'pre_existing' from the DB migration scripts as well?

aim/db/model_base.py
```diff
@@ -65,6 +65,11 @@ class HasTenantName(object):
     tenant_name = name_column(nullable=False)
 
 
+class IsManaged(object):
+    """Add to subclasses that are manager objects"""
```
amittbose: nit: manager -> managed

Also might just want to call this call "IsMonitored" instead because monitored objects are actually not being managed by AIM/AID

aim/db/hashtree_db_listener.py
ivar-lazzaro: A of today, you can't modify relationships of monitored objects (eg: adding a contract to a pre-existing EPG). I'm not sure we even need that, and if we do then we may want to have a separate child objects for those relationships.

aim/db/hashtree_db_listener.py
ivar-lazzaro: They are all created together transactionally, so one cannot exist without the others.

aim/db/migration/alembic_migrations/versions/8e313fbeb93b_l3out.py
ivar-lazzaro: yes, good catch

aim/db/model_base.py
ivar-lazzaro: ok

aim/db/hashtree_db_listener.py
amittbose: I think being able to modify relationship of pre-existing objects is a basic requirement.

aim/db/hashtree_db_listener.py
amittbose: Where is that being enforced?

ivar-lazzaro: It's not enforced by the DB per se, but if you do that through the TreeManager then it will be the case:

https://github.com/noironetworks/aci-integration-module/blob/monitored-universe/aim/db/tree_model.py#L106-L124

aim/db/hashtree_db_listener.py
ivar-lazzaro: We can use monitored objects as a reference, and create children for them, but I'm not sure modifying the references of a monitored object is a basic requirement. What use case do you have in mind? One thing would be possible is to add contracts to pre-existing EPGs, but I'm not sure that's a basic requirement.

amittbose: I believe it is required to support some external connectivity configurations. In any case, the design for monitored objects doesn't seem to account for the possibility that we may want to modify relations of pre-existing objects, and that looks like a major weakness to me.

ivar-lazzaro: The problem with relationships is that in our model they are homogeneous, meaning that at any given point in time is not clear what is and what isn't monitored/managed. We could probably solve the issue by modeling the relationships themselves as an object. thoughts?

amittbose: This issue was already solved in the way I had set up the converter to handle pre-existing objects. The converter decides which relations are managed, and which are not.

ivar-lazzaro:
Look at https://github.com/noironetworks/aci-integration-module/pull/41 instead

PR #40, opened 2016-09-27
Introduce ACI Security Group model

ivar-lazzaro:


ivar-lazzaro:
I still don't have a fabric where I can actually try this fix, but it should be fine to start reviewing it

PR #41, opened 2016-10-07
Feature sync error

ivar-lazzaro:


aim/agent/aid/universes/aci/converter.py
```diff
@@ -287,6 +290,8 @@ def fv_rs_dom_att_converter(object_dict, otype, helper,
 # of the conversion for obtaining the final result
 # - convert_pre_existing: If True, AIM-to-ACI conversion of the object will
 # be performed when the AIM object is marked "pre-existing". Default is False.
+# - convert_monitored: If False, AIM-to-ACI conversion of the object will
```
amittbose: Since attribute 'pre_exisiting' is not a defined attribute in the AIM API, its special use when building the config hash tree and relation to 'monitored' needs to be mentioned somewhere

PR #50, opened 2016-12-08
Support display_name in AIM, translated into APIC nameAlias

ivar-lazzaro:


aim/agent/aid/universes/aci/converter.py
```diff
@@ -720,6 +728,9 @@ def convert(self, aim_objects):
                     continue
                 is_pre = getattr(object, 'pre_existing', False)
                 is_mon = getattr(object, 'monitored', False)
+                if 'display_name' in object.__dict__:
+                    object.__dict__['name_alias'] = object.display_name
```
amittbose: Hopefully we will never have an AIM attribute called "name_alias" or an ACI attribute called "displayName"

PR #51, opened 2016-12-10
Propagate object status to the proper relatives

ivar-lazzaro:
- When sync status goes to FAILED, propagate to all the children
  recursively;
- When sync status goes to PENDING, propagate to parent (unless already
  synced) and children recursively (no object can be created without
  its parent anyways);
- When a new object is created in a subtree, apply the above.

### DIFF
--- a/aim/aim_manager.py
+++ b/aim/aim_manager.py
@@ -73,6 +73,14 @@ class AimManager(object):
                      api_res.ExternalSubnet: models.ExternalSubnet,
                      api_infra.HostLink: infra_model.HostLink}
 
+    # Build adjacency graph (Key: <ACI Resource> Value: <Key's children>)
+    _model_tree = {}
+    for klass in _db_model_map:
+        try:
+            _model_tree.setdefault(klass._tree_parent, []).append(klass)
+        except AttributeError:
+            pass
+
     def __init__(self):
         # TODO(amitbose): initialize anything we need, for example DB stuff
         self._resource_map = {}
@@ -102,8 +110,10 @@ class AimManager(object):
                     attr_val = self._extract_attributes(resource, "other")
                     db_obj.from_attr(context.db_session, attr_val)
             db_obj = db_obj or self._make_db_obj(context.db_session, resource)
-            context.db_session.add(db_obj)
             self._add_commit_hook(context.db_session)
+            context.db_session.add(db_obj)
+            # Propagate sync status to neighbor objects
+            self.set_resource_sync_pending(context, resource)
             return self.get(context, resource)
 
     def update(self, context, resource, **update_attr_val):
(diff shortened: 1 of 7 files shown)


aim/aim_manager.py
```diff
@@ -117,8 +127,6 @@ def update(self, context, resource, **update_attr_val):
         made to the database.
         """
         self._validate_resource_class(resource)
-        if not update_attr_val:
```
amittbose: Was this removed so that the sync status can be set to pending even with a no-op update?

aim/aim_manager.py
```diff
@@ -235,25 +243,68 @@ def update_status(self, context, resource, status):
...
+                # Change parent first
+                parent_klass = resource._tree_parent
+                if parent_klass:
+                    identity = {v: resource.identity[i]
+                                for i, v in enumerate(
+                        parent_klass.identity_attributes)}
+                    # Send an empty update instead of simply recurring this
+                    # method. By doing this, we make sure that the resource
+                    # is re-added to the hash tree
+                    self.set_resource_sync_pending(context,
```
amittbose: code doesn't match the comment above

aim/aim_manager.py
```diff
@@ -73,6 +73,14 @@ class AimManager(object):
                      api_res.ExternalSubnet: models.ExternalSubnet,
                      api_infra.HostLink: infra_model.HostLink}
 
+    # Build adjacency graph (Key: <ACI Resource> Value: <Key's children>)
+    _adjacency_graph = {}
```
amittbose: this is just a tree, not really an adjacency graph

aim/api/status.py
```diff
@@ -58,7 +58,7 @@ class AciStatus(resource.ResourceBase, OperationalResource):
     def __init__(self, **kwargs):
         super(AciStatus, self).__init__({'resource_type': None,
                                          'resource_id': None,
-                                         'sync_status': self.SYNC_PENDING,
+                                         'sync_status': None,
```
amittbose: Not sure I understand the need for this. Is there a real reason for sync_status to be ever unset? It seems useful to AIM users if they can rely on getting a valid sync_status value always.

aim/aim_manager.py
ivar-lazzaro: yes

aim/api/status.py
ivar-lazzaro: By doing this, creating a new object can trigger the neighborhood sync propagation. If this was already in sync pending, the propagation would stop at the current object. Keep in mind that we set an object status to SYNC_PENDING right after its creation, so we will never show an empty state to the user

aim/aim_manager.py
ivar-lazzaro: will change

PR #52, opened 2016-12-14
Only use apic_system_id config option in the infra CLI script

ivar-lazzaro:
Remove dependency from that config everywhere else

### DIFF
--- a/aim/config.py
+++ b/aim/config.py
@@ -32,12 +32,6 @@ from aim import exceptions as exc
 
 LOG = logging.getLogger(__name__)
 
-global_opts = [
-    cfg.StrOpt('apic_system_id',
-               default='openstack',
-               help="Prefix for APIC domain/names/profiles created"),
-]
-
 
 common_opts = [
     cfg.StrOpt('host', default=socket.gethostname(),
@@ -86,6 +80,7 @@ agent_opts = [
                       "situations.")),
 ]
 
+# TODO(ivar): move into AIM section
 event_service_polling_opts = [
     cfg.FloatOpt('service_polling_interval', default=10,
                  help=("Number of seconds or fraction of it that the polling "
@@ -122,7 +117,6 @@ class ConfigManager(object):
             apic_config.apic_opts + apic_config.apic_opts_from_ml2},
         'aim': {
             x.dest: x for x in agent_opts},
-        'default': {x.dest: x for x in global_opts},
         'aim_event_service_polling': {x.dest: x for x in
                                       event_service_polling_opts}
     }
@@ -246,8 +240,7 @@ class ConfigManager(object):
             return
         self.subs_mgr.register_callback(callback, item, group, host, version)
 
-    def get_option_and_subscribe(self, callback, item, group='default',
-                                 host=None):
+    def get_option_and_subscribe(self, callback, item, group='aim', host=None):
         host = host or self.host
         opt = self._get_option(item, group, host)
         self.option_subscribe(callback, item, group, host, opt['version'])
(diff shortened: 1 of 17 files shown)


aim/config.py
```diff
@@ -32,18 +32,6 @@
 
 LOG = logging.getLogger(__name__)
 
-global_opts = [
```
amittbose: Heads up: removing these options will break the AIM ML2 driver (driver initialize and UTs). Please be ready with follow-on patches to the ML2 driver.

ivar-lazzaro: Neutron already provides a "host" global option, this was only registered in cases where the standard Neutron config wasn't available (eg. the CLI or AIM UTs)

aim/tests/base.py
```diff
@@ -28,10 +28,18 @@
 from aim.db import api
 from aim.db import model_base
 
+
+global_opts = [
```
ivar-lazzaro: When the APICAPI config options are loaded (for testing the CLI for example) they need apic_system_id to be globally registered otherwise they will break as a consequence of their inheritance

aim/config.py
amittbose: I was mentioning [1] and [2]

[3] seems to be taking care of [1], though this patch needs work to address review comments. You still need a patch for [2]

[1] http://git.openstack.org/cgit/openstack/group-based-policy/tree/gbpservice/neutron/plugins/ml2plus/drivers/apic_aim/mechanism_driver.py#n107
[2] http://git.openstack.org/cgit/openstack/group-based-policy/tree/gbpservice/neutron/tests/unit/plugins/ml2plus/test_apic_aim.py#n96
[3] https://review.openstack.org/#/c/412022/

aim/config.py
```diff
@@ -84,8 +72,12 @@
     cfg.BoolOpt('recovery_restart', default=True,
                 help=("Set to True if you want the agents to exit in critical "
                       "situations.")),
+    cfg.StrOpt('aim_service_identifier', default=socket.gethostname(),
+               help="(NODB) Identifier for this specific AID service, "
```
amittbose: In this is something that is relevant to the end user, the one is may chaneg this option? Should it be in the help string at all? If you really want to put it there, perhaps something more readable like "Not persisted in database" at the end of the text will be better

etc/aim/aimctl.conf
```diff
@@ -0,0 +1,60 @@
+[DEFAULT]
+apic_system_id=openstack
+
+
+[apic]
+# The following snippet is derived from the APICAPI conf file
+# Hostname:port list of APIC controllers
+# apic_hosts = 1.1.1.1:80, 1.1.1.2:8080, 1.1.1.3:80
```
amittbose: Putting these options in the sample config file is just confusing. I think they should be removed.

aim/config.py
ivar-lazzaro: [0] should now address both cases

[0] https://review.openstack.org/#/c/412022/

aim/config.py
ivar-lazzaro: I think it is relevant for operators that need to know whether a restart is required or not while changing an option. I haven't opted for something more readable because we will have to copy-paste this string for every non persistent option.

etc/aim/aimctl.conf
ivar-lazzaro: ok will do

aim/config.py
amittbose: If the important thing about these options is whether restart is required or not, then the help string should mention that instead of what looks like an implementation detail. Also, as I understand, the help string is for human consumption, so perhaps human readability should be given more weightage.

PR #53, opened 2016-12-21
Set tenant as a foreign key for AIM resources

ivar-lazzaro:
New semantics:

- Tenant object must always exist in order for children to be
  created;
- Tenant object now supports faults;
- When deleting an AIM object, the AIM manager will try to delete
  its whole monitored subtree.

### DIFF
--- a/aim/db/migration/alembic_migrations/versions/3ef7607a41b0_contracts.py
+++ b/aim/db/migration/alembic_migrations/versions/3ef7607a41b0_contracts.py
@@ -42,7 +42,9 @@ def upgrade():
         sa.PrimaryKeyConstraint('aim_id'),
         sa.UniqueConstraint('tenant_name', 'name',
                             name='uniq_aim_filters_identity'),
-        sa.Index('idx_aim_filters_identity', 'tenant_name', 'name'))
+        sa.Index('idx_aim_filters_identity', 'tenant_name', 'name'),
+        sa.ForeignKeyConstraint(
+            ['tenant_name'], ['aim_tenants.name'], name='fk_flt_tn'))
 
     op.create_table(
         'aim_filter_entries',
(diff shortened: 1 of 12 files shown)


aim/db/migration/alembic_migrations/versions/3ef7607a41b0_contracts.py
```diff
@@ -42,7 +42,9 @@ def upgrade():
         sa.PrimaryKeyConstraint('aim_id'),
         sa.UniqueConstraint('tenant_name', 'name',
                             name='uniq_aim_filters_identity'),
-        sa.Index('idx_aim_filters_identity', 'tenant_name', 'name'))
+        sa.Index('idx_aim_filters_identity', 'tenant_name', 'name'),
+        sa.ForeignKeyConstraint(
+            ['tenant_name'], ['aim_tenants.name'], name='fk_tn'))
```
amittbose: I think the FK constraint names have to be unique in MySQL. Can you confirm that using the same name for all constraints is OK?

ivar-lazzaro: I assumed they need to be unique per table, but I can't confirm 100%

amittbose: MySQL doesn't permit same constraint name:

MariaDB [neutron]> create table a(id integer primary key);
Query OK, 0 rows affected (0.72 sec)

MariaDB [neutron]> create table b(id integer, constraint fk foreign key (id) references a(id));
Query OK, 0 rows affected (0.14 sec)

MariaDB [neutron]> create table c(id integer, constraint fk foreign key (id) references a(id));
ERROR 1005 (HY000): Can't create table 'neutron.c' (errno: 121)

PR #63, opened 2017-01-08
Avoid deleting objects when in error state

ivar-lazzaro:
Today, every time ad AIM object gets in SYNC_FAILED state, we
remove the corresponding entry from the proper hashtree. This can
sometimes cause unexpected datapath loss due to the subsequent
deletion of the object itself from APIC.

For example, an ACI VRF would face deletion from APIC just because
an update to its display name set an invalid value! To avoid
such issue, this patch puts the hashtree node in a new state called
"error" instead of deleting it from the tree. When a node is in error
state, it will be excluded from the hashtree difference calculation
without stopping the procedure, so that its subtree can be checked
for inconsistencies.

TODO: error nodes still participate the hash calculation, making the
diff operation more inefficient when such nodes are present.

### DIFF
--- a/aim/api/resource.py
+++ b/aim/api/resource.py
@@ -69,7 +69,7 @@ class ResourceBase(object):
     @property
     def members(self):
         return {x: self.__dict__[x] for x in self.attributes() +
-                ['pre_existing'] if x in self.__dict__}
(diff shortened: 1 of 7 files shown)


amittbose:
Patch is in merge conflict, can you rebase your treee?

PR #78, opened 2017-03-23
Always set classPref in fvRsDomAtt objects

amittbose:
The classPref is set to 'useg' to allow micro-
segmentation EPGs. This should not affect other
EPGs that don't use micro-segmentation.

Also db-migration now does an implicit hash-tree
reset to account for schema and converter changes.

### DIFF
--- a/aim/agent/aid/universes/aci/converter.py
+++ b/aim/agent/aid/universes/aci/converter.py
@@ -253,8 +253,10 @@ def fv_rs_dom_att_converter(object_dict, otype, helper,
             dn = default_identity_converter(
                 object_dict, otype, helper, extra_attributes=[phys_dn],
                 aci_mo_type='fvRsDomAtt', to_aim=False)[0]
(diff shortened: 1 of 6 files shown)


aim/agent/aid/universes/aci/converter.py
```diff
@@ -253,8 +253,10 @@ def fv_rs_dom_att_converter(object_dict, otype, helper,
             dn = default_identity_converter(
                 object_dict, otype, helper, extra_attributes=[phys_dn],
                 aci_mo_type='fvRsDomAtt', to_aim=False)[0]
-            result.append({'fvRsDomAtt': {'attributes': {'dn': dn,
-                                                         'tDn': phys_dn}}})
+            result.append({'fvRsDomAtt': {'attributes':
+                                          {'dn': dn,
+                                           'tDn': phys_dn,
+                                           'classPref': 'useg'}}})
```
ivar-lazzaro: Is this compatible with ACI version 2.0?

PR #82, opened 2017-04-05
Add lockmode 'update' on hashtree select during oncommit operations

ivar-lazzaro:


aim/tree_manager.py
```diff
@@ -59,7 +59,7 @@ def update_bulk(self, context, hash_trees, tree=CONFIG_TREE):
         trees = {self.tenant_rn_funct(x): x for x in hash_trees}
         self._add_commit_hook(context)
         with context.store.begin(subtransactions=True):
-            db_objs = self._find_query(context, tree,
+            db_objs = self._find_query(context, tree, lock_update=True,
```
amittbose: do we need lock_update=True in delete_bulk also?

PR #83, opened 2017-04-05
Add support for using in-cluster config for simpler deployment.

readams:


aim/k8s/api_v1.py
```diff
@@ -101,7 +101,10 @@ def __init__(self, api_client=None, config_file=None):
             self.api_client = api_client
         else:
             if not config.api_client:
-                konfig.load_kube_config(config_file=config_file)
+                if config_file is not None and config_file != "":
```
amittbose: Nit:  "if config_file:" is sufficient to check for both cases

PR #96, opened 2017-04-26
Define VMM injected objects and use native Kubernetes objects when possible

amittbose:


aim/agent/aid/universes/k8s/k8s_watcher.py
```diff
@@ -133,50 +135,98 @@ def _renew_klient_watch(self):
...
+                    kwargs['name'] = self.namespace
+                for event in self.klient.watch.stream(
+                        self.wrap_list_call, k8s_type,
+                        _thread_state=my_state,
+                        namespace=self.namespace):
+                    if my_state['watch_stop']:
+                        LOG.debug('Stopping %s objects thread %s',
+                                  k8s_type.kind, my_thread)
+                        break
+                    LOG.debug("Kubernetes event (%s objects) received: %s",
```
ivar-lazzaro: Is it worth having an info level log with some basic info of the event received? Like kind, name and status?

aim/aim_store.py
```diff
@@ -312,13 +320,25 @@ def _before_session_commit(session, flush_context, instances):
...
+
+    def __init__(self, namespace=None, config_file=None,
+                 vmm_domain=None, vmm_controller=None):
         super(K8sStore, self).__init__()
         self.klient = api_v1.AciContainersV1(config_file=config_file)
         self.namespace = namespace or api_v1.K8S_DEFAULT_NAMESPACE
+        self.attribute_defaults = {'domain_type': 'Kubernetes',
+                                   'domain_name': vmm_domain or 'kubernetes',
+                                   'controller_name':
+                                   vmm_controller or 'kube-cluster'}
```
ivar-lazzaro: should these be configuration driven?

aim/aim_store.py
```diff
@@ -329,60 +349,63 @@ def supports_hooks(self):
...
+                    curr.pop('status', None)
+                    self.klient.replace(k8s_klass, db_obj['metadata']['name'],
+                                        self.namespace, curr)
+                    created = curr
+                break
+            except api_v1.klient.ApiException as e:
+                if str(e.status) == '404':
+                    # Object doesn't exist, create it.
+                    db_obj.pop('resourceVersion', None)
+                    self.klient.create(k8s_klass, self.namespace, db_obj)
```
ivar-lazzaro: 409 could be raised here if the same object is created concurrently. would our intent be to retry if that happens?

aim/aim_store.py
ivar-lazzaro: Seems like they are config driver, then why do we need the default values here?

aim/db/migration/alembic_migrations/versions/5d975a5c2d60_vmm_injected_objects.py
```diff
@@ -0,0 +1,228 @@
+# Copyright (c) 2016 Cisco Systems
```
ivar-lazzaro: nit: 2017

aim/db/migration/alembic_migrations/versions/5d975a5c2d60_vmm_injected_objects.py
```diff
@@ -0,0 +1,228 @@
...
+def upgrade():
+    op.create_table(
+        'aim_vmm_controllers',
+        sa.Column('aim_id', sa.Integer, autoincrement=True),
+        sa.Column('domain_type', sa.String(64), nullable=False),
+        sa.Column('domain_name', sa.String(64), nullable=False),
+        sa.Column('name', sa.String(64), nullable=False),
+        sa.Column('display_name', sa.String(256), nullable=False, default=''),
+        sa.Column('monitored', sa.Boolean, nullable=False, default=False),
+        sa.Column('scope', sa.Enum('unmanaged', 'vm', 'iaas', 'network',
```
ivar-lazzaro: Should we really enumerate the DB rows? Schema already controls it, and invalid values will just be rejected by ACI. A new controller type would require to migrate the database.

aim/db/models.py
```diff
@@ -666,3 +674,221 @@ class Topology(model_base.Base, model_base.AttributeMixin,
...
+
+class VMMController(model_base.Base, model_base.HasDisplayName,
+                    model_base.HasAimId, model_base.AttributeMixin,
+                    model_base.IsMonitored, model_base.HasName):
+    """DB model for VMM Controller."""
+    __tablename__ = 'aim_vmm_controllers'
+    __table_args__ = (
+        model_base.uniq_column(__tablename__, 'domain_type', 'domain_name',
+                               'name') +
+        (sa.ForeignKeyConstraint(
```
ivar-lazzaro: With the AIM-as-a-Schema approach, we should be able to create objects independently of their ACI parents, making the use of FKs an inconsistency between SQL and K8S. Even from a testing perspective, we want the same test suite to be able to run on both systems, which means that we eventually want to remove all FKs. For now, we might think about not adding new ones.

amittbose: I disagree about not adding new ones. It is preferable to maintain consistency with existing code instead of diverging for some cases. If and when we decide to remove FK constraints, it should be trivial to remove them from a few more tables.

aim/db/migration/alembic_migrations/versions/5d975a5c2d60_vmm_injected_objects.py
amittbose: I don't think DB migration is a problem. Generally speaking, whether the backend store should be more restrictive in accepting arbitrary values for enumerations is a topic worth discussing. It is however out of scope of this change because we should fix all enumerations instead of starting a new pattern for these objects.

aim/db/migration/alembic_migrations/versions/5d975a5c2d60_vmm_injected_objects.py
amittbose: ok

aim/aim_store.py
amittbose: I don't think you would get a 409 (Conflict) for a create (HTTP POST), it is usually reported replace (HTTP PUT)

aim/aim_store.py
amittbose: this store doesn't receive the config object, so it cannot rely on get the defaults from the config file. The defaults are provided for the sake of completeness of this class.

aim/agent/aid/universes/k8s/k8s_watcher.py
amittbose: I think logging every kubernetes event will make the log very verbose. Instead I do log an actionable event here: https://github.com/noironetworks/aci-integration-module/pull/96/files#diff-41168944953075e7986281e4c8144702R322. I found this more useful to debug.

amittbose:
Addressed review comments.

PR #106, opened 2017-05-06
Delete fault objects correctly even if parentless

ivar-lazzaro:


aim/agent/aid/universes/aim_universe.py
```diff
@@ -98,55 +98,21 @@ def get_resources(self, resource_keys):
...
-                                                               dissected[0])
-            rns = dn_mgr.filter_rns(mos_rns)
-            # TODO(amitbose) We should be using 'alt_resource' only if we don't
-            # find an AIM object by using 'resource'
-            conv_info = converter.resource_map[aci_klass][0]
-            klass = conv_info.get('alt_resource') or conv_info['resource']
-            res = klass(
-                **dict([(y, rns[x])
-                        for x, y in enumerate(klass.identity_attributes)]))
+        resources = self.get_resources_for_delete(resource_keys)
```
amittbose: Since we will now ignore 'alt_resource' from the converter map, I think there will be problems with sync-ing in cases where there are high/low-level objects.

ivar-lazzaro: I will revert this change as it's not relevant to the main issue. That said, why doesn't this affect the get_resources_for_delete method? Can it be modified so that we only run 1 code path to translate resource keys into AIM objects?
Also, we should add UTs reproducing the issue you are mentioning, or patches similar to this risk to sneak in eventually.

PR #107, opened 2017-05-06
Reset k8s watcher retries number after a clean monitoring iteration

ivar-lazzaro:
when observe_and_monitor_loop runs, it only exits the function when
an error occurs or the thread is stopped. This because it's blocked
in a loop internal to the one already provided by the decorator.
Because of that, whenever an error occurs its retry count is increased
by the decorator without ever having a chance to be reset. This
change makes sure that the only loop in this method is the
decorator itself.

In addition, save hashtrees from the K8S watcher every time for
operational resources, even if they look like empty update, so
that we don't lose information about the node error state.

### DIFF
--- a/aim/agent/aid/universes/k8s/k8s_watcher.py
+++ b/aim/agent/aid/universes/k8s/k8s_watcher.py
@@ -149,29 +149,30 @@ class K8sWatcher(object):
     def persistence_thread(self):
         self._thread(self.persistence_loop, "K8S Tree Builder")
 
-    @utils.retry_loop(OBSERVER_LOOP_MAX_WAIT, OBSERVER_LOOP_MAX_RETRIES,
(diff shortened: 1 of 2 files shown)


aim/agent/aid/universes/k8s/k8s_watcher.py
```diff
@@ -353,7 +357,9 @@ def _process_event(self, event):
                                   aim_ctx=self.ctx)
             new_hash = (cfg.root_full_hash, mo.root_full_hash,
                         oper.root_full_hash)
-            if old_hash != new_hash:
+            # Operational state changes can modify trees without changing
+            # their hash
+            if old_hash != new_hash or is_oper:
```
amittbose: you probably want to add a UT for operational resource events.

PR #128, opened 2017-07-13
Fix URL rotation for websockets

ivar-lazzaro:


mikecohen:
can we put in ebro release too?

PR #129, opened 2017-07-18
Change concrete device interface length

ivar-lazzaro:


ivar-lazzaro:
Is there any place where a similar problem could occur?

aim/db/service_graph_model.py
```diff
@@ -85,7 +85,7 @@ class DeviceClusterInterfaceConcreteIfs(model_base.Base):
     dci_aim_id = sa.Column(sa.Integer,
                            sa.ForeignKey('aim_device_cluster_ifs.aim_id'),
                            primary_key=True)
-    interface = model_base.name_column(primary_key=True)
+    interface = model_base.dn_column(primary_key=True)
```
amittbose: Instead of sa.String(512), this needs to specified as VARCHAR(512, charset='latin1') because MySQL restricts the byte length of primary-key columns.

PR #133, opened 2017-09-12
HTTP REST server for AIM

ivar-lazzaro:
A new process 'aim-http-server' is introduced to expose a simple REST
API that wraps AIM functionalities and stores the AIM model into
the programmed store.
New configuration options have been added:

[aim_server]
port=8080
socket_file=/var/lib/aim/server.sock
binding_ip=127.0.0.1


amittbose:
Change looks good. You'll need to also update the debian packaging scripts for the new server: add .conf and .service file in debian/ directory and then reference these files in debian/install

PR #136, opened 2017-10-12
Add the HostLinkNetworkLabel model and corresponding DB table.

kentwu1:


aim/api/infra.py
```diff
@@ -81,3 +81,16 @@ def __init__(self, **kwargs):
         super(HostDomainMapping, self).__init__({'vmm_domain_name': '',
                                                  'physical_domain_name': ''},
                                                 **kwargs)
+
+
+class PhysnetIFMapping(resource.ResourceBase):
```
amittbose: As far as AIM/ACI is concerned, "physical-network" doesn't mean anything. This is more like adding a "label" to the host interface so that OpenStack can interpret it a particular way. I suggest calling this something like HostLinkNetworkLabel and rename 'physical_network_name' to 'network_label'

kentwu1:
pushed a revision to address the review comments.

PR #138, opened 2017-10-23
Add config option to disable micro-segmentation

tbachman:
Some situations, such as older switch hardware, require us to have
the ability to disable configuration of Micro-Segmentation support
on the VMM domain associations with EPGs. This patch adds that
configuration option, which removes this attribute from APIC
configuration when associating VMM Domains with EPGs.

### DIFF
--- a/aim/agent/aid/universes/aci/converter.py
+++ b/aim/agent/aid/universes/aci/converter.py
@@ -27,6 +27,7 @@ from aim.api import resource
 from aim.api import status as aim_status
 from aim.api import types as t
 from aim.common import utils as aim_utils
+from aim import config as aim_cfg
 
 LOG = logging.getLogger(__name__)
 DELETED_STATUS = "deleted"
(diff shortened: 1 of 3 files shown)


aim/agent/aid/universes/aci/converter.py
```diff
@@ -98,7 +99,8 @@ def fault_inst_to_resource(converted, helper, to_aim=True):
 
 def fv_rs_dom_att_converter(object_dict, otype, helper,
                             source_identity_attributes,
-                            destination_identity_attributes, to_aim=True):
+                            destination_identity_attributes, to_aim=True,
+                            disable_micro_segmentation=False):
```
ivar-lazzaro: This variable is unused here, as the method is reading directly from the config file.

tbachman: Good catch (this was left over after the initial attempt at passing this in)! Will remove and update the patch.

tbachman:
Pushed an update to the patch for review.

PR #140, opened 2017-10-27
Change default value of BD IP Learning

tbachman:
The default value of the "Limit IP Learning to Subnet" flag
was False when creating DBs. This can lead to problems where
IPs become unreachable in deployments. This patch changes the
default value of this flag to True.


tbachman:
This will be fixed in the GBP code instead (don't want to change the semantics of the BridgeDomain resource). Closing.

PR #142, opened 2017-10-31
Generalize HostDomainMapping table

tbachman:
The vmm_domain_name column is changed to just domain_name and made
a primary key, while the physical_domain_name is dropped in favor
of a domain_type, which is an enum for the supported Domain types.
This is needed in order to allow multiple entries with the same
hostname in the table, which can be used to support a catch-all
or "wildcard" of all hosts for multiple VMM domains.


aim/db/infra_model.py
```diff
@@ -131,9 +131,11 @@ class HostDomainMapping(model_base.Base, model_base.AttributeMixin):
     __tablename__ = 'aim_host_domain_mapping'
 
     host_name = sa.Column(sa.String(128), primary_key=True)
-
-    vmm_domain_name = sa.Column(sa.String(64))
-    physical_domain_name = sa.Column(sa.String(64))
+    domain_name = sa.Column(sa.String(64), primary_key=True, nullable=True)
+    domain_type = sa.Column(sa.Enum('PhysDom',
```
kentwu1: We also need to make domain_type a primary key also right? Otherwise how do you handle the case that a host belongs to 2 domains with the same name but of different type (say physdom and openStack) ?

aim/db/infra_model.py
```diff
@@ -131,9 +131,11 @@ class HostDomainMapping(model_base.Base, model_base.AttributeMixin):
     __tablename__ = 'aim_host_domain_mapping'
 
     host_name = sa.Column(sa.String(128), primary_key=True)
-
-    vmm_domain_name = sa.Column(sa.String(64))
-    physical_domain_name = sa.Column(sa.String(64))
+    domain_name = sa.Column(sa.String(64), primary_key=True, nullable=True)
```
kentwu1: it doesn't make sense to me that domain_name can be Null.....

aim/db/migration/alembic_migrations/versions/5e29ae4f45e6_host_domain_add_primary_key.py
```diff
@@ -0,0 +1,97 @@
...
+    op.drop_table(AIM_HOST_DOMAIN_MAPPING)
+    op.create_table(
+        AIM_HOST_DOMAIN_MAPPING,
+        sa.Column('host_name', sa.String(128)),
+        sa.Column('domain_name', sa.String(64)),
+        sa.Column('domain_type', sa.Enum('PhysDom',
+                                         'OpenStack',
+                                         'Kubernetes',
+                                         'VMware')),
+        sa.PrimaryKeyConstraint('host_name', 'domain_name')
```
kentwu1: all 3 fields should be the primary key I think.

aim/db/infra_model.py
tbachman: This was left-over from some other attempt at the patch -- will remove.

aim/db/infra_model.py
tbachman: Yes, you are correct. Will fix.

aim/db/migration/alembic_migrations/versions/5e29ae4f45e6_host_domain_add_primary_key.py
tbachman: Will fix (as above).

tbachman:
Pushed update to pull-request, addressing reviewer comments.

aim/tests/unit/test_aim_manager.py
```diff
@@ -952,15 +952,16 @@ class TestHostLinkMixin(object):
...
-                                'physical_domain_name': 'physdom1'}
+                                'domain_name': 'ostack1',
+                                'domain_type': 'OpenStack'}
     test_search_attributes = {'host_name': 'host1.example.com'}
-    test_update_attributes = {'vmm_domain_name': 'ostack2',
-                              'physical_domain_name': 'physdom2'}
-    test_default_values = {'vmm_domain_name': '',
-                           'physical_domain_name': ''}
+    test_update_attributes = {}
+    test_default_values = {'domain_name': '',
```
kentwu1: I don't see these 2 default values are being set anywhere in the code so wonder how could this UT pass? I think it makes sense to set domain_type default value to openStack but it does not makes sense to set the default value for domain_name as '' though. Because if we do that, then this entry basically becomes useless. So maybe just set the default value for domain_type then.

tbachman: I'm thinking that the default for the domain_name should be our wildcard ("*").

tbachman:
Posted update, addressing reviewer comments.

aim/api/infra.py
```diff
@@ -76,14 +76,17 @@ class HostDomainMapping(resource.ResourceBase):
...
+        ('domain_type', t.enum('PhysDom',
+                               'OpenStack',
+                               'Kubernetes',
+                               'VMware')))
+    other_attributes = t.other()
 
     def __init__(self, **kwargs):
-        super(HostDomainMapping, self).__init__({'vmm_domain_name': '',
-                                                 'physical_domain_name': ''},
+        super(HostDomainMapping, self).__init__({'domain_name': '*',
```
kentwu1: The '*' should be in the host_name field instead though. But even if you do that, I think the "aimctl manager" command is not going to work well as its expecting 3 parameters so it doesn't know which one is which one say if you just provide 2 parameters there.....Chatted with Amit a bit about this. He said if you really want to provide a default value for the primary key parameters then you also have to modify the CLI parser code which is not trivial. So here I will suggest not to provide any default values for all these parameters as they are all primary key.

tbachman: Thanks for finding that out! Pushed an update that reverts/removes the defaults, and fixes UT issue.

aim/tests/unit/test_aim_manager.py
```diff
@@ -952,15 +952,16 @@ class TestHostLinkMixin(object):
...
-                                'physical_domain_name': 'physdom1'}
+                                'domain_name': 'ostack1',
+                                'domain_type': 'OpenStack'}
     test_search_attributes = {'host_name': 'host1.example.com'}
-    test_update_attributes = {'vmm_domain_name': 'ostack2',
-                              'physical_domain_name': 'physdom2'}
-    test_default_values = {'vmm_domain_name': '',
-                           'physical_domain_name': ''}
+    test_update_attributes = {}
+    test_default_values = {'domain_name': '*',
```
kentwu1: this should be removed. Not sure why it would pass the test though.

tbachman:
Updated pull request to address reviewer comments. I've also had a chance to test the DB migration which worked. The one question outstanding is whether it makes sense to use an enum for the domain_type parameter -- the CLI currently doesn't provide a way of listing valid values for required parameters. A string could be used, but my preference is to keep the data typing, and create the ability to list valid values for parameters in a separate patch. We also should have a better way of testing the REST (i.e. other than assuming all fields of an object accepts a UUID). Again, this should be addressed in a separate patch.

PR #146, opened 2017-11-06
Update External BD entries

tbachman:
Commit fc9ee7b1794986c18e8a024b0b9c4653f82709da changed the
defualt value for the "Limit IP learning to subnets" flag. A
migration is needed to update existing data in the table.

### DIFF
--- a/aim/db/migration/alembic_migrations/versions/dd2f91cf1b1e_set_nat_bd_ip_learn.py
+++ b/aim/db/migration/alembic_migrations/versions/dd2f91cf1b1e_set_nat_bd_ip_learn.py
@@ -1,0 +1,81 @@
+# Copyright (c) 2017 Cisco Systems
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+"""Fix default value for BD's Limit IP Learning to subnets
+
+Revision ID: dd2f91cf1b1e
+Revises: 5e29ae4f45e6
+Create Date: 2017-11-06 16:39:33.713695
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = 'dd2f91cf1b1e'
+down_revision = '5e29ae4f45e6'
+branch_labels = None
+depends_on = None
+testing = True
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.sql.expression import update
+
+AIM_BRIDGE_DOMAINS = 'aim_bridge_domains'
+AIM_BRIDGE_DOMAINS_L3OUTS = 'aim_bridge_domain_l3outs'
+
+# A model of the BridgeDomainL3Out table with the fields we're interested
+bd_l3outs = sa.Table(AIM_BRIDGE_DOMAINS_L3OUTS, sa.MetaData(),
+                     sa.Column('bd_aim_id',
+                               sa.Integer,
+                               sa.ForeignKey('aim_bridge_domains.aim_id'),
+                               primary_key=True),
+                     sa.Column('name', sa.String(64)))
+
+# A model of the BridgeDomain table with the fields we're interested
+bridge_domains = sa.Table(AIM_BRIDGE_DOMAINS, sa.MetaData(),
+                          sa.Column('name', sa.String(64)),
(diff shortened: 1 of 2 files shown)


aim/db/migration/alembic_migrations/versions/dd2f91cf1b1e_set_nat_bd_ip_learn.py
```diff
@@ -0,0 +1,82 @@
...
+                                    primary_key=True, autoincrement=True),
+                          sa.Column('limit_ip_learn_to_subnets',
+                                    sa.Boolean))
+
+
+def upgrade():
+    session = sa.orm.Session(bind=op.get_bind())
+    my_bd_l3outs = get_bd_l3out_values(session)
+    for bd_l3out in my_bd_l3outs:
+        bd_update = update(bridge_domains).values(
```
amittbose: I wonder if this update could be done as a join between the BD table and BD-L3out table ... or possibly with a IN query - SQLAlchemy equivalent of
UPDATE aim_bridge_domain SET limit_ip_learn_to_subnets = 1 WHERE aim_id IN (SELECT DISTINCT bd_aim_id FROM aim_bridge_domain_l3outs)

aim/db/migration/alembic_migrations/versions/dd2f91cf1b1e_set_nat_bd_ip_learn.py
```diff
@@ -0,0 +1,82 @@
...
+            {'limit_ip_learn_to_subnets': True}).where(
+                bridge_domains.c.aim_id == bd_l3out['bd_aim_id'])
+        session.execute(bd_update)
+
+
+def get_bd_l3out_values(session):
+    values = []
+    for row in session.query(bd_l3outs).all():
+        bd_id, name = row
+        values.append({'name': name,
```
amittbose: The name column is not used

aim/db/migration/alembic_migrations/versions/dd2f91cf1b1e_set_nat_bd_ip_learn.py
tbachman: Just trying to make sure I understand what you're optimizing here -- are you thinking this can be done with a single update clause (i.e. without iterating over the bd_l3out entries)?

tbachman: Oh -- I think I see what you're saying. Rather than collect and iterate over the entries from the aim_bridge_domain_l3outs table, just create the aim_bridge_domains row updates from a single query via the join?

amittbose:
Yeah, that's what I meant - use join or IN query to update all the relevant
rows in a single SQL statement.

On Nov 13, 2017 7:06 PM, "Thomas Bachman" <notifications@github.com> wrote:

*@tbachman* commented on this pull request.
------------------------------

In aim/db/migration/alembic_migrations/versions/dd2f91cf1b1e_set_nat_bd_ip_
learn.py
<https://github.com/noironetworks/aci-integration-module/pull/146#discussion_r150727213>
:

> +                          sa.Column('tenant_name', sa.String(64)),
+                          sa.Column('display_name', sa.String(64)),
+                          sa.Column('monitored', sa.Boolean,
+                                    nullable=False, default=False),
+                          sa.Column('aim_id',
+                                    sa.Integer,
+                                    primary_key=True, autoincrement=True),
+                          sa.Column('limit_ip_learn_to_subnets',
+                                    sa.Boolean))
+
+
+def upgrade():
+    session = sa.orm.Session(bind=op.get_bind())
+    my_bd_l3outs = get_bd_l3out_values(session)
+    for bd_l3out in my_bd_l3outs:
+        bd_update = update(bridge_domains).values(

Oh -- I think I see what you're saying. Rather than collect and iterate
over the entries from the aim_bridge_domain_l3outs table, just create the
aim_bridge_domains row updates from a single query via the join?

—
You are receiving this because your review was requested.

Reply to this email directly, view it on GitHub
<https://github.com/noironetworks/aci-integration-module/pull/146#discussion_r150727213>,
or mute the thread

aim/db/migration/alembic_migrations/versions/dd2f91cf1b1e_set_nat_bd_ip_learn.py
tbachman: I tried creating the above, but I run into problems with the Foreign Key. I tried fixing this by adding the relationship to the BridgeDomain table definition, but had trouble getting it to work. If this patch wasn't just for DB migrations (i.e. normal operational code path), I'd be inclined to spend more time figuring out how to make this work, but given that it's just for DB upgrades (and works), I'd prefer to leave this as a potential future optimization for now (maybe add a REVISIT?).

aim/db/migration/alembic_migrations/versions/dd2f91cf1b1e_set_nat_bd_ip_learn.py
tbachman: good catch. Will remove.

tbachman:
Posted an update to the pull-request, addressing one of the reviewer comments. The other issue was responded to.

PR #147, opened 2017-11-09
New HostDomainMapping object

tbachman:
The previous HostDomainMapping object is being deprecated,
in favor of a new one (HostDomainMappingV2), which adds two
primary keys to the table, and therefore has a different API
signature.


aim/db/migration/alembic_migrations/versions/5e29ae4f45e6_host_domain_add_primary_key.py
```diff
@@ -0,0 +1,95 @@
...
+    op.bulk_insert(domainsv2, migrations)
+    # we can get clear out the old table
+    delete_all = delete(domains)
+    session.execute(delete_all)
+
+
+def get_values(session, domains):
+    """Get pre-migration DB rows.
+
+    This gets the rows from the DB before the migration, using
```
kentwu1: this comment is no longer true with this new v2 approach.

aim/db/migration/alembic_migrations/versions/5e29ae4f45e6_host_domain_add_primary_key.py
```diff
@@ -0,0 +1,95 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.sql.expression import delete
+
+AIM_HOST_DOMAIN_MAPPING = 'aim_host_domain_mapping'
+AIM_HOST_DOMAIN_MAPPING_V2 = 'aim_host_domain_mapping_v2'
+
+# A model of the old domains table
+domains = sa.Table(AIM_HOST_DOMAIN_MAPPING, sa.MetaData(),
```
kentwu1: do we still need this since the old table still exists in this approach?

aim/db/migration/alembic_migrations/versions/5e29ae4f45e6_host_domain_add_primary_key.py
```diff
@@ -0,0 +1,95 @@
...
+        sa.Column('domain_name', sa.String(64)),
+        sa.Column('domain_type', sa.Enum('PhysDom',
+                                         'OpenStack',
+                                         'Kubernetes',
+                                         'VMware')),
+        sa.PrimaryKeyConstraint('host_name', 'domain_name', 'domain_type')
+    )
+
+    op.bulk_insert(domainsv2, migrations)
+    # we can get clear out the old table
```
kentwu1: nit: don't need the 'get' in this comment I think

aim/db/migration/alembic_migrations/versions/5e29ae4f45e6_host_domain_add_primary_key.py
```diff
@@ -0,0 +1,95 @@
...
+                   sa.Column('vmm_domain_name', sa.String(64)),
+                   sa.Column('physical_domain_name', sa.String(64)))
+
+testing = True
+
+
+def upgrade():
+    session = sa.orm.Session(bind=op.get_bind())
+    if testing:
+        op.bulk_insert(domains, [{'host_name': 'host1',
```
kentwu1: why do we need this?

aim/db/migration/alembic_migrations/versions/5e29ae4f45e6_host_domain_add_primary_key.py
```diff
@@ -0,0 +1,95 @@
+# Copyright (c) 2017 Cisco Systems
```
kentwu1: should make the file name more clear like 5e29ae4f45e6_host_domains_mapping_v2.py as we are no longer adding primary keys to the old table.

aim/db/migration/alembic_migrations/versions/5e29ae4f45e6_host_domain_add_primary_key.py
tbachman: heh -- something used in testing that I forgot to remove :-). Will remove. Thanks!

aim/db/migration/alembic_migrations/versions/5e29ae4f45e6_host_domain_add_primary_key.py
tbachman: good point -- will rename.

aim/db/migration/alembic_migrations/versions/5e29ae4f45e6_host_domain_add_primary_key.py
tbachman: You're right :-). Will remove.

aim/db/migration/alembic_migrations/versions/5e29ae4f45e6_host_domain_add_primary_key.py
tbachman: You're correct. Will remove.

aim/db/migration/alembic_migrations/versions/5e29ae4f45e6_host_domain_add_primary_key.py
kentwu1: I got that but why do u have to spell out all the details of this table? I will think the table name aim_host_domain_mapping itself should suffice, right?

tbachman: You need to declare the columns you want to access -- if you just declare the table name, it will assume it has no columns.

kentwu1: the point here is that this is the existing table and in this approach we are not modifying this table so we should just import the model class then re-use that instead of creating the same model again locally.

tbachman:
Pushed update to pull-request, addressing reviewer comments.

aim/db/migration/alembic_migrations/versions/5e29ae4f45e6_host_domains_mapping_v2.py
```diff
@@ -0,0 +1,88 @@
...
+            mgr.delete(ctx, mapping)
+
+
+def get_values():
+    additions = []
+    raw_objects = []
+    mgr = aim_manager.AimManager()
+    ctx = context.AimContext(db_session=api.get_session(expire_on_commit=True))
+    with ctx.db_session.begin(subtransactions=True):
+        for mapping in mgr.find(ctx, infra.HostDomainMapping):
```
kentwu1: we can save the return value of mgr.find(ctx, infra.HostDomainMapping) before the transaction starts then later on you don't have to do raw_objects.append() to build it up again. Also I think once you have built up the additions list, at the end of the transaction after the for loop is done, you can just delete the saved returned value. This way we can get rid of delete_old_data() API completely.

tbachman: good advice -- I'll probably keep that method, as I still need to build up a session for the delete.

tbachman: Just provided another update, addressing reviewer comments.

PR #148, opened 2017-11-10
Allow manual provided/consumed contracts on managed EPGs

ivar-lazzaro:
Specify a new converter attribute "skip_for_managed" which intructs
the ACI tenant to not convert this object when untagged and
belonging to a managed parent.


snaiksat:
Tested with basic ICMP and TCP contracts, and seems to be working as expected (Groups created via OpenStack, Contracts created and provided/consumed via APIC).

PR #153, opened 2017-12-05
Monitored universe should not create Root objects

ivar-lazzaro:
There are some cases where contention between the AIM Monitored
Universe and a user happen during Tenant manipulation. In these
cases, the user might delete a monitored Tenant while the universe
was going to update it, thus ending up with a stale tenant even
though the first deletion actually succeeded. This happens because
the Monitored universe overwrites objects instead of updating them,
we should make sure that root objects call the manager "update"
function instead.


tbachman:
Verified with test on FAB8.

PR #155, opened 2017-12-19
Define resources and converters for SVI

kentwu1:
This includes node profile, interface profile and static routes, ...etc.
Have defined only those resources needed for SVI though so for example
routed-subinterface is not covered here. BGP related attributes are also
not covered in this patch.


ivar-lazzaro:
For example, look at [0]. In here, the DeviceCluster object (if enough information is provided) is converted in a huge list of ACI objects, even vnsLIf that could also result from the conversion of the low level model DeviceClusterInterface [1].
Would something similar make sense here, so that Neutron's configuration can be simplified for our use case?

[0] https://github.com/noironetworks/aci-integration-module/commit/983d5e677949454f3880049e4587f3c49bd53349#diff-ceacb4cc3a6a8efbc87370b9473a1104R1849
[1] https://github.com/noironetworks/aci-integration-module/commit/983d5e677949454f3880049e4587f3c49bd53349#diff-c8ed95720f5ddc39857099337f992635R72

kentwu1:
Just discussed with Amit. And we both think we can combine the StaticRoute with the NextHop. The reason I made it a separate class is because a static route could have multiple next_hops but I think its doable to combine both of them.

kentwu1:
Pushed a revision to address the review comments.

PR #158, opened 2018-01-05
Change tenant removal algorithm.

ivar-lazzaro:
The previous algorithm for tenant removal was proved to be unreliable
in situations of high load, where there are lots of threads and race
conditions are likely. This would often cause leftover objects laying
around, since the monitored universe would keep syncing in the background
without causing the removal to cease.

With the new algorithm, we have a two phase decision that is taken
before and after the reconciliation cycle. Furthermore, we make sure
that the monitored universes delete all the AIM monitored objects
when they agree on tenant removal. We do so by pretending that the
desired state is empty. Only when a tenant's AIM state is completely
empty (all trees with no root) we finally cleanup the state.
DB transactionality and update lock should ensure this happens
reliably.

### DIFF
--- a/aim/agent/aid/service.py
+++ b/aim/agent/aid/service.py
@@ -86,11 +86,6 @@ class AID(object):
              CURRENT: aim_universe.AimDbMonitoredUniverse().initialize(
                  api.get_store(), self.conf_manager, self.multiverse)},
         ]
-        # delete_candidates contains tenants that are candidate for deletion.
-        # when the consensus is reach by all the universe, the state will
-        # be cleaned up by the reconcile action
-        self.delete_candidates = {}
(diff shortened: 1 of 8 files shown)


ivar-lazzaro:
Please do not merge yet, but reviews are welcome

tbachman:
I tested this functionally -- I did a full tempest run which created a ton of tenants, then used the aimctl manager tenant-delete command with --force --cascade. It removed all of the tenants and the objects below them. Unless Amit has any objections, I think this PR can be merged (note: it may need to be rebased).

ivar-lazzaro:
Thomas: were the tenants also removed on APIC?

tbachman:
Yes.

PR #161, opened 2018-01-16
Sync Flag, Service Graph conversion fix, SG linear nodes order fix.

ivar-lazzaro:

### DIFF
--- a/aim/db/migration/alembic_migrations/versions/bca1ef645fe1_epg_sync_flag.py
+++ b/aim/db/migration/alembic_migrations/versions/bca1ef645fe1_epg_sync_flag.py
@@ -1,0 +1,43 @@
+# Copyright (c) 2018 Cisco Systems
+# All Rights Reserved.
+#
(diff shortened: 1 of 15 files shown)


aim/db/migration/alembic_migrations/versions/bca1ef645fe1_epg_sync_flag.py
```diff
@@ -0,0 +1,51 @@
...
+import sqlalchemy as sa
+
+
+def upgrade():
+
+    op.add_column(
+        'aim_endpoint_groups',
+        sa.Column('sync', sa.Boolean, server_default=sa.literal(True))
+    )
+    op.add_column(
```
kentwu1: since we are adding this sequence_number also in this migration script, we should probably also enhance the file naming to reflect this change. Also I don't see the HEAD file being changed though.....

ivar-lazzaro:
changed

PR #165, opened 2018-01-27
Code housekeeping

ivar-lazzaro:
- Work around the ACI bugs that doesn't return VMMP root object
  upon subscription. Fake it, so that AID can show the smiley face
  again;
- Remove some debug logs that were flooding the loggin file.


amittbose:
Change looks fine, can you fix the UT failures?

PR #168, opened 2018-02-01
Also handle the case when l3extIp__Member comes.

kentwu1:
This will happen while doing ACI to AIM conversion for the secondary
IPs of an SVI for a VPC setup.

### DIFF
--- a/aim/tests/unit/agent/aid_universes/test_converter.py
+++ b/aim/tests/unit/agent/aid_universes/test_converter.py
@@ -911,6 +911,9 @@ class TestAciToAimConverterL3OutInterface(TestAciToAimConverterBase,
          'resource': 'l3extIp',
          'converter': converter.l3ext_ip_converter},
         {'exceptions': {},
+         'resource': 'l3extIp__Member',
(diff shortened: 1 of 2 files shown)


aim/tests/unit/agent/aid_universes/test_converter.py
```diff
@@ -910,6 +910,9 @@ class TestAciToAimConverterL3OutInterface(TestAciToAimConverterBase,
         {'exceptions': {},
          'resource': 'l3extIp',
```
ivar-lazzaro: it doesn't seem right that the reverse map converter shows both l3extIp and l3extIp__Member.

Reverse map is responsible for AIM to ACI conversion. every L3OutInterface will end up being converted generating both a l3extIp and a l3extIp__Member objects. Then the squash function will make sure only one is left (as it's based on the object's DN). Should this map only have l3extIp__Member instead?

ivar-lazzaro: I see now that the converter has 2 diffrent paths depending on the object type, and that you actually need both unlike Kiran's case. Ignore my comment

PR #170, opened 2018-02-02
Use annotation instead of Tags for ownership

ivar-lazzaro:
Starting F, APIC introduced the concept of annotation, designed
to let orchestrators declare ownership over managed objects.
The APIC team wants us to use annotation instead of tags because
the latter impact performance greately, both in terms of memory
and CPU cycle (a task is generated for every new tag).

This patch want to keep compatibility with APIC back and forth,
so we first test APIC for annotation support and then start using
it for ownership, falling back to tags if support is missing.
On upgrade, I expect AIM to just suddenly forget about the
existing tags and push annotations around. This might require an
external script to get rid of the hanging tags.


ivar-lazzaro:
Do not merge yet

ivar-lazzaro:
Could we review this without merging at the moment?

tbachman:
Abandoned, as it's covered by pull request #267

PR #177, opened 2018-02-13
Drop local_asn parameter

shastrinator:
As part of bgp peering, local_asn as a parameter is not commonly required,
hence dropping the column.


kentwu1:
I got that its not commonly required but it shouldn't hurt to keep it there in case user needs it some day, right? I mean we already provide a good default value so user doesn't have to do anything with it if they don't want to. Is there a reason we have to drop it for sure?

PR #178, opened 2018-02-14
Add load-mappings command

tbachman:
The load-mappings command is added, so that the information
specified in the configuration file can be used to create
entries in the HostDomainMappingV2 table. This patch also
adds a call to this command in the load-domains command,
so that the mappings can be created when the domains first
get loaded.

### DIFF
--- a/aim/tools/cli/commands/manager.py
+++ b/aim/tools/cli/commands/manager.py
@@ -22,6 +22,7 @@ from oslo_log import log as logging
 from tabulate import tabulate
 
 from aim import aim_manager
+from aim.api import infra
 from aim.api import resource
 from aim.api import schema
 from aim.api import status as status_res
@@ -206,11 +207,34 @@ def describe(klass):
     return _describe
 
 
+def do_mappings(aim_ctx, manager, replace):
+    if replace:
+        curr_mappings = manager.find(aim_ctx, infra.HostDomainMappingV2)
+
+        for mapping in curr_mappings:
+            click.echo("Deleting %s: %s" % (type(mapping), mapping.__dict__))
+            manager.delete(aim_ctx, mapping)
+
+    vmm_doms = manager.find(aim_ctx, resource.VMMDomain)
+    phys_doms = manager.find(aim_ctx, resource.PhysicalDomain)
+    doms = vmm_doms + phys_doms
+    for dom in doms:
+        if isinstance(dom, resource.PhysicalDomain):
+            domtype = 'PhysDom'
+        else:
+            domtype = dom.type
+        res = infra.HostDomainMappingV2(host_name=infra.WILDCARD_HOST,
+                                        domain_type=domtype,
+                                        domain_name=dom.name)
+        print_resource(manager.create(aim_ctx, res, overwrite=True))
+
+
 @manager.command(name='load-domains')
 @click.option('--replace/--no-replace', default=False)
 @click.option('--enforce/--no-enforce', default=False)
+@click.option('--mappings/--no-mappings', default=True)
 @click.pass_context
-def load_domains(ctx, replace, enforce):
+def load_domains(ctx, replace, enforce, mappings):
     manager = ctx.obj['manager']
     aim_ctx = ctx.obj['aim_ctx']
     vmm_types = utils.KNOWN_VMM_TYPES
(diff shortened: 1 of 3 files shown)


tbachman:
Hold off on merging this -- I want to check one thing first.

tbachman:
Minor update -- is now open for review.

aim/tools/cli/commands/manager.py
```diff
@@ -240,6 +264,8 @@ def load_domains(ctx, replace, enforce):
             res = resource.PhysicalDomain(name=phys, monitored=True)
             print_resource(manager.create(aim_ctx, res, overwrite=True))
 
+        if mappings:
+            do_mappings(aim_ctx, manager, replace)
```
amittbose: I don't think that --replace with load_domains should carry over to do_mappings - we probably don't want to accidentally erase any manually created host-domain mapping entries when user updates the config file domain information.

aim/tools/cli/commands/manager.py
```diff
@@ -248,12 +274,42 @@ def load_domains(ctx, replace, enforce):
                           for x in manager.find(aim_ctx,
                                                 resource.PhysicalDomain)]
             all_epgs = manager.find(aim_ctx, resource.EndpointGroup)
+            if mappings:
+                host_mappings = {}
```
amittbose: Nit: do we need this dict? Two variables would be just fine

aim/tools/cli/commands/manager.py
```diff
@@ -248,12 +274,42 @@ def load_domains(ctx, replace, enforce):
...
+                # We use just the wildcard mappings, since we aren't asking
+                # APIC for host-to-EPG associations
+                db_mappings = manager.find(aim_ctx, infra.HostDomainMappingV2,
```
amittbose: I'm a little confused about this block: do_mappings() earlier in the function creates wildcarded entries for all resource.VMMDomain and resource.PhysicalDomain objects. And then this block queries the domains associated with all wildcarded entries. Won't the result be the same as what the existing code was doing (lines 271-275)?

kentwu1: I'm confused here also......

aim/tools/cli/commands/manager.py
tbachman: I think there's still a use case for blowing away the old mappings. I could make this a separate parameter?

aim/tools/cli/commands/manager.py
tbachman: It is confusing :-(. I probably should explain this better. If the user doesn't use replace, then this will create entries from the domains it finds, but there *could* also be entries that the user had created previously in the HostDomainMappingV2 table (i.e. from running the aimctl manager host-domain-mapping-v2-create). Would additional commenting be helpful?

aim/tools/cli/commands/manager.py
tbachman: Sure -- will change.

aim/tools/cli/commands/manager.py
kentwu1: I think some comment here would definitely help!

amittbose: OK. Say if the user had previously created non-wildcard entries in HostDomainMappingV2, then you are not considering those at all, and you will end up removing the association between existing EPGs and those domains, right? I think the correct thing to do here is *append* to the list of EPG-domains associations instead of replacing that list.

aim/tools/cli/commands/manager.py
amittbose: Blowing away old mappings can be done through the new command (load-mappings). Do you need to duplicate that there?

tbachman: Sounds good. Will change.

aim/tools/cli/commands/manager.py
tbachman: Good point.

The 90% use case won't involve the --replace (probably only needed if a domain is no longer needed and must be removed, in which case the user will also want the host mapping removed, so they will have to know to remove it explicitly). It's mainly for folks who want to define an initial set of domains for an openstack instance, and then possibly to add to it. If there are host-specific mappings, then they will have to add those using the separate command anyway, so I think I agree with your points. Will change.

tbachman: I just realized that the --enforce is a bit more complicated. I think it really only can be useful when there are only wildcard mappings. If you have a mix of host-specific and wildcard mappings, you can't make associations because you have no information about hosts (neutron has port-to-host mapping information -- you have no way to know whether a host-specific domain mapping or a wildcard domain mapping should be used). My thinking here is that the enforce option would ignore host-specific mappings and assume that a new set of mappings is being done, using only wildcard mappings.

amittbose: May be --enforce should error out when there are host-specific mappings instead of potentially breaking existing EPGs?

tbachman: Yeah -- either that, or I could check if the EPG already has a domain, and if there is a host-specific mapping with that domain, then don't update that EPG. That said, I think the erroring out makes more sense, and would cover the 90% use case workflow. The host-specific mapping can only be associated through port-binding, so it doesn't make sense to try to do something clever here, IMHO.

tbachman:
Just pushed an update for review, with the new behavior.

PR #189, opened 2018-03-02
Recover session as a result of a SQLAlchemyError

ivar-lazzaro:


ivar-lazzaro:
In case we decide to merge this, I guess we can squash the commits.

tbachman:
Tested this as a regression of the east-west GBP datapath tests, and it worked. Recommend merging.

PR #191, opened 2018-03-06
Make aimctl consistent with neutron's -list and -show

ivar-lazzaro:


aim/tests/unit/tools/cli/test_manager.py
```diff
@@ -439,10 +439,10 @@ def update(self, res_command, attributes):
         return self._parse(res)
 
     def find(self, res_command, attributes):
-        return self._run_manager_command(res_command, 'find', attributes)
+        return self._run_manager_command(res_command, 'list', attributes)
```
amittbose: Please test both variations of the commands.

PR #194, opened 2018-03-13
Introduce SQL optimizer

ivar-lazzaro:
In some cases we might want to use more complicated queries than
the ones allowed by the aim manager for improved performance.
For such cases, introduce a sql optimizer module which exposes
some specific search methods and tries to resolve them with
the least number of sqlalchemy queries possible.

### DIFF
--- a/aim/db/migration/alembic_migrations/versions/HEAD
+++ b/aim/db/migration/alembic_migrations/versions/HEAD
@@ -1,1 +1,1 @@
-f1ca776aafab
+1b58ffa871bb
--- a/aim/tests/unit/test_utils.py
+++ b/aim/tests/unit/test_utils.py
@@ -22,6 +22,7 @@ Tests for `utils` module.
 
 import mock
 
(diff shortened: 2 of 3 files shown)


aim/aim_lib/db/optimizer.py
```diff
@@ -0,0 +1,50 @@
...
+            if diff:
+                raise WrongStoreType(features=requirements)
+            else:
+                return func(*args, **kwargs)
+        return inner
+    return wrap
+
+
+@requires(['sql'])
+def get_epg_by_host_names(context, host_names):
```
amittbose: This could just be a function in aim/utils that has a fast implementation for SQL stores and a default brute force one for other stores. That way the AIM library users don't have to worry about special cases.
Also, the name of this function should perhaps be something like 'get_epg_by_static_path_hosts'.

ivar-lazzaro:
Do not merge for now.

PR #196, opened 2018-03-17
Version control

ivar-lazzaro:


aim/aim_store.py
```diff
@@ -397,6 +395,8 @@ def _before_session_commit(session, flush_context, instances):
                     (session.deleted, deleted)]
         for mod_set, res_list in modified:
             for db_obj in mod_set:
+                if isinstance(db_obj, model_base.AttributeMixin):
+                    db_obj.bump_epoch()
```
amittbose: I think a comment is needed here explaining that concurrent modification of the underlying DB object will cause SQLAlchemy error and hence rollback of the transaction.
Also can we add a UT to simulate concurrent update? For example, register a wrapper sa_event callback that commits a change in a new session and then invokes the callback registered by the store?

aim/db/model_base.py
```diff
@@ -99,6 +95,21 @@ class IsSynced(object):
 class AttributeMixin(object):
     """Mixin class for translating between resource and model."""
 
+    epoch = sa.Column(
+        sa.BigInteger().with_variant(sa.Integer(), 'sqlite'),
+        server_default='0', nullable=False)
```
amittbose: In the DB migration script, nullable is set to True. Is this discrepancy needed?

ivar-lazzaro: Nope, will fix

aim/aim_store.py
ivar-lazzaro: I'll add a link to the official sqlalchemy doc.

I could simulate a transaction with 2 AIM operations on a EPG and a version bump occurring in the middle by a second session.

PR #197, opened 2018-03-20
Add host column to L3OutInterface objects

ivar-lazzaro:


aim/db/migration/data_migration/add_host_column.py
```diff
@@ -74,3 +89,9 @@ def migrate(session):
             session.execute(update(DeviceClusterDevice).where(
                 DeviceClusterDevice.c.path == hlink.path).values(
                 host=hlink.host_name))
+            session.execute(update(DeviceClusterDevice).where(
```
amittbose: is this update() same as the previous statement?

PR #203, opened 2018-03-30
Introduce model for IPSLA

ivar-lazzaro:
- ServiceRedirectMonitoringPolicy
- ServiceRedirectHealthGroup
- references from ServiceRedirectPolicy


aim/db/migration/alembic_migrations/versions/7ba6b6346710_ip_sla_tables.py
```diff
@@ -0,0 +1,84 @@
+# Copyright (c) 2017 Cisco Systems
```
amittbose: 2018

aim/agent/aid/universes/aci/converter.py
```diff
@@ -994,7 +995,16 @@ def _squash(self, converted_list):
                               enumerate(klass.identity_attributes)])))
             for k, v in res.__dict__.iteritems():
                 if isinstance(v, list):
-                    current.__dict__.setdefault(k, []).extend(v)
+                    for item in v:
+                        if isinstance(item, dict) and '_converter_id' in item:
+                            id_ = (klass.__name__, ) + item.pop(
```
amittbose: id_ should also include k, the AIM resource attribute

ivar-lazzaro: good catch!

PR #218, opened 2018-04-18
Introduce sync state recover command

ivar-lazzaro:
Which will update all the objects in error state for easier
recovery procedure.

### DIFF
--- a/aim/tools/cli/commands/manager.py
+++ b/aim/tools/cli/commands/manager.py
@@ -377,6 +377,32 @@ def sync_state_find(ctx, state, plain):
                         tablefmt='plain' if plain else 'psql'))
 
 
+@manager.command(name='sync-state-recover')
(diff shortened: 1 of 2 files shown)


aim/tools/cli/commands/manager.py
```diff
@@ -377,6 +377,32 @@ def sync_state_find(ctx, state, plain):
                         tablefmt='plain' if plain else 'psql'))
 
 
+@manager.command(name='sync-state-recover')
+@click.pass_context
+def sync_state_recover(ctx):
+    manager = ctx.obj['manager']
+    aim_ctx = ctx.obj['aim_ctx']
+    error = [status_res.AciStatus.SYNC_FAILED, 'error', 'sync_error', 'failed']
```
tbachman: nit: better as a (immutable) tuple?

PR #220, opened 2018-04-23
Change AIM ID column from increment to UUID

ivar-lazzaro:
Looks like sqlalchemy re-uses incremental IDs whenever objects are
deleted from the DB. This, together with the fact that there are
cases in which status objects leak without a parent, can cause the
system to go in a non recoverable state where the Status object
of a new AIM resource clashes with a zombie status of the same type.

Although zombie status periodic cleanup needs to be addressed, this
patch only concerns about avoiding the status clash from happeing.

### DIFF
--- a/aim/db/hashtree_db_listener.py
+++ b/aim/db/hashtree_db_listener.py
@@ -19,6 +19,7 @@ from oslo_log import log as logging
 from oslo_utils import importutils
 
 from aim.api import resource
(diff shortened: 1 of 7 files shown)


ivar-lazzaro:
do not merge yet

ivar-lazzaro:
Reviews/Merging open now

Issue #231, opened 2018-07-26
Critical DB Error due to constraints in PostGreSQL Join-handing. 


mmnelemane:
"aimctl db-migration upgrade head" results in the following error when used against a postgresql backend.  This is due to the inherent way postgres handles left outer joins.

The reason for this behavior from postgres is explained in the mailing list here: https://www.postgresql.org/message-id/10878.1131978292%40sss.pgh.pa.us

CRITI [aim] DBError: (psycopg2.NotSupportedError) FOR UPDATE cannot be applied to the nullable side of an outer join
 [SQL: 'SELECT aim_tenant_trees.tenant_rn AS aim_tenant_trees_tenant_rn, aim_tenant_trees.epoch AS aim_tenant_trees_epoch, aim_tenant_trees.needs_reset AS aim_tenant_trees_needs_reset, aim_agent_to_tree_associations_1.tree_tenant_rn AS aim_agent_to_tree_associations_1_tree_tenant_rn, aim_agent_to_tree_associations_1.agent_id AS aim_agent_to_tree_associations_1_agent_id \nFROM aim_tenant_trees LEFT OUTER JOIN aim_agent_to_tree_associations AS aim_agent_to_tree_associations_1 ON aim_tenant_trees.tenant_rn = aim_agent_to_tree_associations_1.tree_tenant_rn FOR UPDATE']
Traceback (most recent call last):
  File "/usr/bin/aimctl", line 10, in <module>
    sys.exit(aimctl())
  File "/usr/lib/python2.7/site-packages/aim/tools/services/cli.py", line 21, in aimctl
    shell.run()
  File "/usr/lib/python2.7/site-packages/aim/tools/cli/shell.py", line 27, in run
    aim(auto_envvar_prefix='AIM')
  File "/usr/lib/python2.7/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 889, in invoke
[...]

PR #232, opened 2018-07-26
Fix for critical DB error with postgres backend

mmnelemane:
Fixes: https://github.com/noironetworks/aci-integration-module/issues/231

The commit updates the tree_model in migration to use lazy_load (http://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html#lazy-loading) which emits a SELECT statement instead of the Joined Eager Loading (http://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html#joined-eager-loading ) which emits a LEFT OUTER JOIN by default on the SELECT statement causing the postgres backend to produce the error.

### DIFF
--- a/aim/db/tree_model.py
+++ b/aim/db/tree_model.py
@@ -46,7 +46,7 @@ class Tree(model_base.Base, model_base.AttributeMixin):
     agents = orm.relationship(AgentToHashTreeAssociation,
                               backref='hash_trees',
                               cascade='all, delete-orphan',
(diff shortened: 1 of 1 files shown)


ivar-lazzaro:
Could subquery be used instead to keep the same eager loading semantics?

mmnelemane:
@ivar-lazzaro : Do you have any other suggestion. Can we merge this ?

PR #233, opened 2018-07-26
Fix for duplicate uniq_agents

mmnelemane:
The names uniq_agents0agent_type0host conflicts with the unique values created by neutron db schema. This conflict throws errors due to the uniqueness constraint. This commit fixes the unique name constraint with a separate name for aim_agents.

### DIFF
--- a/aim/db/migration/alembic_migrations/versions/accfe645090a_agent_resource.py
+++ b/aim/db/migration/alembic_migrations/versions/accfe645090a_agent_resource.py
@@ -48,7 +48,7 @@ def upgrade():
         sa.Column('version', sa.String(10)),
         sa.PrimaryKeyConstraint('id'),
         sa.UniqueConstraint('agent_type', 'host',
-                            name='uniq_agents0agent_type0host'))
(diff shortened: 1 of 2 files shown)


aim/db/migration/alembic_migrations/versions/accfe645090a_agent_resource.py
```diff
@@ -48,7 +48,7 @@ def upgrade():
         sa.Column('version', sa.String(10)),
         sa.PrimaryKeyConstraint('id'),
         sa.UniqueConstraint('agent_type', 'host',
-                            name='uniq_agents0agent_type0host'))
+                            name='aim_agents_uniq_agents0agent_type0host'))
```
ivar-lazzaro: Changing an existing migration file is not gonna work, as this code won't run on systems that already have AIM installed.
Please revert this file, and create a new migration HEAD that modifies the existing constraint name.

If this makes it impossible to install AIM in any particular system, then keep this file modified. But also add a new HEAD which checks the constraint name and fixes it if required

Issue #230, opened 2018-07-18
DB Migration fails due to possible race in session handling


mmnelemane:
"aimctl db-migration upgrade head" fails at reading aim_vmm_domains table. The error is as below:

INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade  -> 40855b7eb958, Create tables for Tenant and BridgeDomain
INFO  [alembic.runtime.migration] Running upgrade 40855b7eb958 -> 72fa5bce100b, Create TenantTree table
INFO  [alembic.runtime.migration] Running upgrade 72fa5bce100b -> accfe645090a, Create Agent table
INFO  [alembic.runtime.migration] Running upgrade accfe645090a -> 7349fa0a2ad8, Create table for subnet, vrf, app-profile and EPG.
INFO  [alembic.runtime.migration] Running upgrade 7349fa0a2ad8 -> faade1155a0a, Create Status and Fault model
INFO  [alembic.runtime.migration] Running upgrade faade1155a0a -> 3ef7607a41b0, Create tables for contracts and filters
INFO  [alembic.runtime.migration] Running upgrade 3ef7607a41b0 -> 74f15b6aee51, Create endpoint table
INFO  [alembic.runtime.migration] Running upgrade 74f15b6aee51 -> ab9b4e196100, Create Config Model
INFO  [alembic.runtime.migration] Running upgrade ab9b4e196100 -> 1249face3348, Create tables for VMM and Physical Domains.
INFO  [alembic.runtime.migration] Running upgrade 1249face3348 -> 8e313fbeb93b, Tables for L3out
INFO  [alembic.runtime.migration] Running upgrade 8e313fbeb93b -> 07113feba145, Tables for L3out clones in AIM LIB
INFO  [alembic.runtime.migration] Running upgrade 07113feba145 -> d8abdb89ba66, Create table for HostLinks and EPG static-paths.
INFO  [alembic.runtime.migration] Running upgrade d8abdb89ba66 -> d38c07b36c11, Table for saved L3Out in AIM lib
INFO  [alembic.runtime.migration] Running upgrade d38c07b36c11 -> abf7bb5a4100, Create tables for security groups
INFO  [alembic.runtime.migration] Running upgrade abf7bb5a4100 -> babbefa38870, Add conntrack column
[...]

PR #235, opened 2018-07-26
Fix for #230 Error due to non-existing table

mmnelemane:
Fixes: https://github.com/noironetworks/aci-integration-module/issues/230

The session handling is covered under try...except so that when the migration is done on a clean setup without pre-existing tables, it can pass through and not throw errors due to non-existing tables. If it is a pre-existing setup, then the tables will exist and the control goes through the try... section.

### DIFF
--- a/aim/db/migration/alembic_migrations/versions/aceb1ac13668_vmm_policy.py
+++ b/aim/db/migration/alembic_migrations/versions/aceb1ac13668_vmm_policy.py
@@ -36,6 +36,8 @@ from aim.api import resource
 from aim import context
 from aim.db import api
 
+from oslo_db.exception import DBError
+
 
 def upgrade():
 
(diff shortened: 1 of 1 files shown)


aim/db/migration/alembic_migrations/versions/aceb1ac13668_vmm_policy.py
```diff
@@ -59,12 +61,16 @@ def upgrade():
...
+                new_vmms.append(resource.VMMDomain(type=vmm.type, name=vmm.name,
+                                                   monitored=True))
+            for phys in session.query(old_phys_table).all():
+                new_phys.append(resource.PhysicalDomain(name=phys.name,
+                                                        monitored=True))
+        except DBError:
+            # We can fail if the tables are not yet there
+            pass
 
     op.drop_table('aim_vmm_domains')
```
ivar-lazzaro: wouldn't this fail too if the table didn't exist? Also, why wouldn't the table exist? It is created by a previous revision (1249face3348)

mmnelemane: The bug has two aspects:

Problem 1.
As described in the Issue #230, there is a new connection created by aceb1ac13668 before the commit of the earlier transactions started by the previous session. Since these transactions were not committed, the SQL statement creating the table did not commit too. You can refer to the SQL output attached here: https://github.com/noironetworks/aci-integration-module/files/2206403/postgresql.log-201806211158.txt. 

Problem 2. 
Looking at the logic in aceb1ac13668_vmm_policy.py, it is trying to fetch data from the previous schema and populating into the new schema. This logic could work well for existing table. In fact, the above issue (1) does not affect operation since a table from previous deployment already exists. But for a fresh deployment, the old schema tables do not exist and will be created fresh during migration which is not happening due to (1). 

Solution:
The solution in this commit is to ignore failure to read the old tables. If tables do not exist, the queries will result in a DBError exception which we catch and ignore.

Hope this explanation helps.

PR #237, opened 2018-08-24
Avoid sending the delete requests of children to APIC

kentwu1:
If a parent is also getting deleted then we don't have to send
those children delete requests to APIC because APIC can just
remove the whole tree based on the parent's DN.


mandeepdhami:
LGTM

PR #240, opened 2018-09-27
Support indexed utf-8 encoded columns

tbachman:
Some database backends are constrained to a size of 767 bytes for
indexed columns. Many backends use utf-8 for character encoding, which
can use a maximum of 3 bytes per character. This triggers warnings
or errors when running DB migrations. This patch reduces these columns
to a size of 255 bytes. Note that if utf8mb4 character encoding is
used, then these as well as other columns must be reduced to 191 bytes
(and would be fixed by a separate patch), as it uses 4 bytes per character


aim/db/migration/alembic_migrations/versions/2b48bf7f19cb_fix_host_column.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def get_inspector():
+    """Reuse inspector"""
+
+    global _INSPECTOR
+
+    if _INSPECTOR:
+        return _INSPECTOR
+
```
kentwu1: nit: don't need a space here

aim/db/migration/alembic_migrations/versions/1b58ffa871bb_add_host_column.py
```diff
@@ -36,19 +36,19 @@
 def upgrade():
     op.add_column(
         'aim_endpoint_group_static_paths',
-        sa.Column('host', sa.String(512), nullable=True, index=True)
+        sa.Column('host', sa.String(256), nullable=True, index=True)
```
kentwu1: should be 255 across all the changes.

kentwu1: Also there is another migration script needs to be changed I think.

tbachman: Good catches -- will fix both.

tbachman: Just pushed changes to address the reviewer comments.

aim/db/migration/alembic_migrations/versions/fabed2911290_modify_host_column_size.py
```diff
@@ -29,20 +29,46 @@
...
+    return inspector.get_columns(table)
+
+
 def upgrade():
     for table in TABLES:
-        with op.batch_alter_table(table) as batch_op:
-            batch_op.alter_column(
-                'host', existing_type=sa.String(length=1024),
-                type_=sa.String(512))
+        columns = get_columns(table)
```
kentwu1: do we really need to change this file though because the new migration script should be able to cover that?

tbachman: I verified with breakpoints that it will set it, regardless of the length of the string type above, so it is needed.

PR #244, opened 2018-10-09
Get SecurityGroupRule from DB when processing action log

rkukura:


rkukura:
This change uses the current SecurityGroupRule from the DB instead of the JSON-encoded SGR sent with the action log. This should ensure that concurrent updates to the remote_ips attribute of an SGR, which no longer bump the SGR's epoch and therefore are not serialized, result in the correct final state in the AID hash tree. The code is exercised by existing UTs, and I am looking at whether and how to add an additional UT that tests the case where the DB state differs from the JSON state in the action log.

aim/db/hashtree_db_listener.py
```diff
@@ -202,6 +202,33 @@ def _preprocess_logs(self, logs):
...
+                                 "action" % db_aim_res)
+                    else:
+                        # Use current resource from DB so that list
+                        # attributes do no need to be protected from
+                        # concurrent updates by bumping the resource's
+                        # epoch.
+                        aim_res = db_aim_res
+                else:
+                    if action != aim_tree.ActionLog.DELETE:
+                        LOG.warn("AIM resource %s does not exist in DB for "
```
kentwu1: The warning log will only make sense if its for the UPDATE action I think. Because for CREATE action, it makes sense that the record is not in the DB yet.

aim/db/hashtree_db_listener.py
```diff
@@ -202,6 +202,33 @@ def _preprocess_logs(self, logs):
...
+                if db_aim_res:
+                    if action == aim_tree.ActionLog.DELETE:
+                        LOG.warn("AIM resource %s exists in DB for delete "
+                                 "action" % db_aim_res)
+                    else:
+                        # Use current resource from DB so that list
+                        # attributes do no need to be protected from
+                        # concurrent updates by bumping the resource's
+                        # epoch.
+                        aim_res = db_aim_res
```
kentwu1: This should only be done for the UPDATE action I think since adding/removing remote_ips list is considered an UPDATE action to the SGR. But I guess there is no harm to do it for the CREATE action either so it should be ok.

PR #255, opened 2019-04-19
Fix DB migrations

tbachman:
The "Revises" and "Revision ID" entries were incorrect in many
of the migrations. This patch fixes them.


tbachman:
FYI - I came across this while debugging a DB migration issue in the field. We technically don't need this pull-request, as DB migrations are working without it, but it wouldn't hurt to have them all properly linked.

PR #259, opened 2019-06-02
Fix ASN data type

tbachman:
The BGP AS Number size was specified as an Integer, which
supports signed numbers from -2,147,483,648 to 2,147,483,647.
However, BGP ASNs use unsigned 32-bit numbers, which range
from 0 to 4,294,967,295. This patch changes the column from
an Integer to a BigInteger, so that the full ASN range can
be supported.

Fixes #886


aim/db/models.py
```diff
@@ -1100,7 +1100,7 @@ class L3OutInterfaceBgpPeerP(model_base.Base, model_base.HasAimId,
     interface_profile_name = model_base.name_column(nullable=False)
     interface_path = sa.Column(VARCHAR(512, charset='latin1'), nullable=False)
     addr = sa.Column(sa.String(64), nullable=False)
-    asn = sa.Column(sa.Integer)
+    asn = sa.Column(sa.BigInteger)
```
shastrinator: I guess this should be 
from sqlalchemy.dialects.mysql import INTEGER
sa.Column(INTEGER(unsigned=True))
The test seems to be failing expecting a float, not sure why bigint is treated as a float.

tbachman: Hmmm - I may have to rethink this then. I don't think we can rely on dialects, as that's backend-dependent. My guess is I'll have to change the type to a string, and will need to create a data migration to support it (and change the validator), but I'll see if there are any other options. Thanks!

tbachman: Looks like I was wrong - the problem was in the exception definition. Please see the updated pull-request.

PR #261, opened 2019-07-30
Fix query_statuses to return empty list if no aim res provided

skandasw:
Add a check to return back an empty list if no aim resources
are provided to get corresponding status objects.


kentwu1:
The change looks good and I assume this patch shouldn't break any existing GBP UTs right?

tbachman:
Looks like even AIM has an issue: https://travis-ci.org/noironetworks/aci-integration-module/jobs/565696041

Sridar, can you have a look?

PR #264, opened 2019-10-02
Fix query_statuses to return empty list if no aim res provided

skandasw:
Add a check to return back an empty list if no aim resources
are provided to get corresponding status objects.


skandasw:
Verified gbpservice repo with a dummy review with test-requirements set to aim branch containing this change.

https://review.opendev.org/#/c/686078/

PR #265, opened 2019-10-04
Distribute APICs evenly among those aim-aid controllers

kentwu1:
1. Use a new DB table to coordinate the APIC assignments
to make sure they are distributed evenly.

2. When the assigned APIC goes down, aim-aid will try
to find the next available APIC or fall back to one
APIC which has already been assigned to somebody else.
For latter case, it will then enter the recovery mode
and try to find the next available APIC periodically
with a back-off timer.


aim/agent/aid/universes/aci/aci_universe.py
```diff
@@ -113,37 +124,148 @@ def _reload_websocket_config(self):
...
+                    # This means the original aim-aid owner might have
+                    # crashed or something. We will just take it!
+                    if apic_assign_obj and self.session.ipaddr in url:
+                        obj = self._update_apic_assign_db(
+                            aim_context, apic_assign, apic_assign_obj)
+                        if obj is None:
+                            continue
+                        self.need_recovery = False
+                        self.apic_assign_obj = obj
+                        return
```
tbachman: Don't we still need a websocket connection in this case?

aim/agent/aid/universes/aci/aci_universe.py
```diff
@@ -26,13 +26,19 @@
 from aim.agent.aid.universes import base_universe as base
 from aim.agent.aid.universes import constants as lcon
 from aim.agent.aid.universes import errors
+from aim.api import infra as api_infra
 from aim.api import resource
+# from aim.common import hashring
```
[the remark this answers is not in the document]
kentwu1: yes will do.

aim/agent/aid/universes/aci/aci_universe.py
```diff
@@ -113,37 +124,148 @@ def _reload_websocket_config(self):
...
+                # automatically
+                obj = self.manager.update(aim_context, apic_assign_obj)
+            return obj
+        # This means another controller is also adding/updating
+        # the same entry at the same time and he has beat us.
+        except Exception as e:
+            LOG.info(e)
+            return None
+
+    def _ws_session_login(self, url, url_max_reties, purpose):
```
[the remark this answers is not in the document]
kentwu1: yes will fix.

aim/agent/aid/universes/aci/aci_universe.py
```diff
@@ -113,37 +124,148 @@ def _reload_websocket_config(self):
...
+                    apic_assign = api_infra.ApicAssignment(apic_host=url)
+                    apic_assign_obj = self.manager.get(aim_context,
+                                                       apic_assign)
+                    if (apic_assign_obj and
+                        apic_assign_obj.aim_aid_id != self.agent_id and
+                            not apic_assign_obj.is_available(aim_context)):
+                        continue
+                    # This means the original aim-aid owner might have
+                    # crashed or something. We will just take it!
+                    if apic_assign_obj and self.session.ipaddr in url:
```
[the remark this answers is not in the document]
kentwu1: good point. Let me think about it then. Worst case I just don't do this special thing. Then it will always try to re-connect.

aim/agent/aid/universes/aci/aci_universe.py
kentwu1: In this case, the APIC picked for recovery is the same APIC in self.session so no need to re-connect.

aim/agent/aid/universes/aci/aci_universe.py
```diff
@@ -113,37 +124,148 @@ def _reload_websocket_config(self):
...
-                        verify_ssl=self.verify_ssl_certificate,
-                        cert_name=self.cert_name, key=self.private_key_file)
-                    resp = self.session.login()
-                    if not resp.ok:
-                        LOG.warn('Websocket connection failed: %s' % resp.text)
-                        self.ws_urls.rotate(-1)
-                        LOG.info('Rotating websocket URL, '
-                                 'using: %s' % self.ws_urls[0])
-                        retries += 1
+                url_max_reties = max(1, max_retries / len(self.ws_urls))
```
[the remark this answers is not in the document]
kentwu1: yes will fix.

aim/agent/aid/universes/aci/aci_universe.py
```diff
@@ -113,37 +124,148 @@ def _reload_websocket_config(self):
...
+                for url in backup_urls:
+                    session = self._ws_session_login(url, url_max_reties,
+                                                     BACKUP_PURPOSE)
+                    if session:
+                        if self.session and self.session.session:
+                            self.session.close()
+                            self.is_session_reconnected = True
+                        self.session = session
+                        self._spawn_monitors()
+                        self.need_recovery = True
```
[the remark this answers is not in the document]
kentwu1: I can randomly rotate the backup_urls list so that all the extra aim-aids will not go to the same APIC in the backup_urls.

aim/aim_manager.py
```diff
@@ -98,7 +100,6 @@ class AimManager(object):
                      api_service_graph.DeviceClusterInterfaceContext,
                      api_service_graph.ServiceRedirectMonitoringPolicy,
                      api_service_graph.ServiceRedirectHealthGroup,
-                     api_infra.OpflexDevice,
```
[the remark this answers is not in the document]
kentwu1: oh it was just being moved to line 82 so that all the classes under api_infra are grouped together. Whoever added it didn't follow the convention so.....

aim/aim_store.py
```diff
@@ -263,7 +265,6 @@ class SqlAlchemyStore(AimStore):
                         service_graph_model.ServiceRedirectMonitoringPolicy),
                     api_service_graph.ServiceRedirectHealthGroup: (
                         service_graph_model.ServiceRedirectHealthGroup),
-                    api_infra.OpflexDevice: infra_model.OpflexDevice,
```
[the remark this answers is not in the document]
kentwu1: ditto

aim/agent/aid/universes/aci/aci_universe.py
```diff
@@ -263,10 +413,10 @@ def _thread_monitor(self, flag):
             utils.perform_harakiri(LOG, msg)
 
 
-def get_websocket_context(apic_config):
+def get_websocket_context(apic_config, aim_manager):
     global ws_context
     if not ws_context:
-        ws_context = WebSocketContext(apic_config)
+        ws_context = WebSocketContext(apic_config, aim_manager)
```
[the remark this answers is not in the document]
kentwu1: I can certainly add a REVISIT. Or if u prefer, for now I can initialize a new aim_manager member variable inside WebSocketContext. I passed it in in this patch because aci_universe already has an aim_manager member variable available so.

aim/agent/aid/universes/aci/aci_universe.py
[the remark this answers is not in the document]
kentwu1: By examining the code, actually this will be ok coz self.session.ipaddr could also be DNS name. The code basically just strips out the "https://" from the beginning then assign it to ipaddr member variable. Agreed that ipaddr naming is confusing somehow.....
https://github.com/noironetworks/acitoolkit/blob/noiro-lite/acitoolkit/acisession.py#L503-L506

aim/agent/aid/universes/aci/aci_universe.py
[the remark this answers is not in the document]
tbachman: I think the REVISIT is good for now - passing explicitly is probably best here.

aim/agent/aid/universes/aci/aci_universe.py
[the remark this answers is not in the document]
tbachman: I just want to make sure we end up with a good distribution. I doubt most people will scale this way, but I don't want to penalize anyone who does. That said, I don't have a good suggestion of how to deal with this.

aim/agent/aid/universes/aci/aci_universe.py
tbachman: got it.

aim/agent/aid/universes/aci/aci_universe.py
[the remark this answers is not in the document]
tbachman: ah, good point. That works then.

aim/agent/aid/universes/aci/aci_universe.py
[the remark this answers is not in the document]
kentwu1: Given the uncommon use case, I think just by randomly rotate the backup_urls list at the beginning is a good workaround to distribute it even further.

kentwu1:
Pushed a revision to address those review comments.

aim/agent/aid/universes/aci/aci_universe.py
```diff
@@ -113,37 +124,152 @@ def _reload_websocket_config(self):
...
+                if session.session:
+                    session.close()
+                continue
+            LOG.info('%s Websocket connection succeeded with url: %s',
+                     purpose, url)
+            return session
+
+        return None
+
+    def establish_ws_session_for_recovery(self, max_retries=None):
```
rkukura: This function looks a lot like establish_ws_session() below. Could they be folded into a single function that handles both "normal" and "recovery" situations? Also, I'm still not really clear on why recovery needs to be handled differently.

aim/db/infra_model.py
```diff
@@ -154,3 +155,12 @@ class HostLinkNetworkLabel(model_base.Base, model_base.AttributeMixin):
     host_name = sa.Column(sa.String(128), primary_key=True)
     network_label = sa.Column(sa.String(64), primary_key=True)
     interface_name = sa.Column(sa.String(32), primary_key=True)
+
+
+class ApicAssignment(model_base.Base, model_base.AttributeMixin):
+    __tablename__ = 'aim_apic_assignment'
+
+    apic_host = sa.Column(sa.String(128), primary_key=True)
+    aim_aid_id = sa.Column(sa.String(64))
```
rkukura: Have you considered also making aim_aid_id a primary key? This would allow the situation where there are more AIDs than APICs to be represented in the DB. 

Then, when connecting, an AID could query for all the records, and pick the APIC that appears to be least loaded. Of course race conditions could result in all AIDs picking the same APIC, so some sort of re-check after connecting might make sense, or some sort of random periodic rebalancing.

aim/agent/aid/universes/aci/aci_universe.py
```diff
@@ -113,37 +124,152 @@ def _reload_websocket_config(self):
...
+                                self.session.close()
+                                self.is_session_reconnected = True
+                            self.session = session
+                            self._spawn_monitors()
+                            self.need_recovery = False
+                            self.apic_assign_obj = obj
+                            return
+                        else:
+                            session.close()
+                            continue
```
rkukura: It seems that similar sequences of code are repeated several times. Should this be refactored into another helper function, or maybe moved into _ws_session_login() itself? 

Its also not really clear why _update_apic_assign_db() is called in some cases but not all. This could be parameterized if needed.

rkukura:
Also, it might be worth splitting the hash ring bug fix (and associated UT) out as a separate patch.

aim/agent/aid/universes/aci/aci_universe.py
kentwu1: The design in this patch is that we use the new table to indicate who has claimed the ownership of an APIC already. If an APIC is already claimed, aim-aid will try to look for the next APIC on which nobody has laid the claim yet. If it couldn't find such an APIC, it will then go ahead to use 1 APIC which already has an owner then enter the recovery mode. When in the recovery mode, this aim-aid will try to find the next free APIC periodically with a back-off timer. Recovery needs to be handled slightly different because if this aim-aid still couldn't locate a free APIC, its totally fine (because currently its already sharing an APIC with some other owner) so no need to re-connect if that happens.

kentwu1: I was actually thinking about using one single function to do it but didn't proceed further as I think code will get a bit messy with quite a few if-else statement here & there. In general I like the idea of avoiding code duplication so I could take another look then.

aim/agent/aid/universes/aci/aci_universe.py
kentwu1: I think it makes sense to put some of that into _ws_session_login() so will look into it. We don't need to call _update_apic_assign_db() when we try to find a back-up APIC because at this point we are going to share an APIC who already has an owner so no need to claim the ownership of that APIC.

aim/db/infra_model.py
kentwu1: As explained in one of the earlier comment, the design in this patch is that we use the new table to indicate who has claimed the ownership of an APIC already instead of showing who is using which APIC so.

kentwu1:
pushed a revision to address those review comments.

aim/agent/aid/universes/aci/aci_universe.py
rkukura: Assuming that the "full-mesh" approach is not workable or would require too much effort to implement, I guess the fundamental concern I have is whether a single AID having "ownership" of an APIC is the right concept to implement. This makes sense only if we assume there are always at least as many APICs as there are AIDs in a healthy deployment. Then, as currently implemented in this patch, the normal situation is for at most one AID to "own" each APIC, and temporary loss of one or more APICs can be thought of as "recovery".

I understand ACI requires at least 3 APICs, and that 3 APICs is a typical deployment configuration. So, is deploying more than 3 AIDs useful? If deploying multiple AIDs is intended only for availability, then 3 AIDs is probably sufficient. If 3 AIDs isn't considered sufficient redundancy, then 3 APICs probably isn't sufficient either, so requiring at least as many APICs as AIDs may be reasonable. But if additional AIDs are needed for scalability, and not just for availability, then more AIDs than APICs becomes a normal situation we should support. This is where the concept of at most one AID owning an APIC becomes problematic. 

If we do need to optimally support configurations with more AIDs than APICs, we could do this by replacing the concept of "ownership" here with one of "association". The DB table would use the AID identity instead of the APIC identity as the primary key, and a row would record which APIC each AID is currently using. Each AID would need to periodically update its row in the table so timestamps would allow rows that have not recently been updated to be ignored. When connecting, an AID would query the whole table, and then randomly pick from the APICs used by the least number of AIDs. I don't think this "association" approach would require significantly more effort to implement than the "ownership" approach used in this patch, but I could be missing something.

If we are convinced that we definitely do not need to (optimally) support configurations with more AIDs than APICs, then my concern with this patch was the code duplication between "normal" and "recovery" cases. The normal case code, _establish_ws_connection(), would be executed all the time, and any bugs in it would be apparent and get fixed. But  _establish_ws_connection_for_recovery() would rarely be executed in production. Any future fixes or enhancements to one would need to be manually applied to the other, or they'd diverge over time. I'll take closer look to see if this has been adequately addressed, or if additional refactoring might be helpful.

aim/agent/aid/universes/aci/aci_universe.py
```diff
@@ -113,37 +124,135 @@ def _reload_websocket_config(self):
...
+                     purpose, url)
+            valid_session = session
+            break
+
+        if valid_session:
+            # We don't need to claim the ownership if we are just
+            # picking up a backup APIC.
+            if purpose != BACKUP_PURPOSE:
+                obj = self._update_apic_assign_db(
+                    aim_context, apic_assign, apic_assign_obj)
```
rkukura: What if two different AIDs concurrently pick the same APIC URL, establish connections, and get to this point? Will the create() call in _update_apic_assign_db() fail for the 2nd one, or will the 2nd create() overwrite the 1st? Seems we need to make sure this AID actually ends up owning the APIC, and if it doesn't, pick a different URL and try again.

kentwu1: In that case, the DB operation inside _update_apic_assign_db() will throw an exception which will be caught then None will be returned for the 2nd AID. Then it will go ahead to try the next URL.

aim/agent/aid/universes/aci/aci_universe.py
kentwu1: My understanding is that most production deployments would have 3 openStack controllers and 3 APICs. In some productions we could even have more APICs up to 7. So its actually rare to have more openStack controllers than APICs. But even if that happens, this patch can handle that well. The extra AIDs will go into recovery mode with back-off timer which eventually will only try to look for free APIC in every 10 mins so it won't cause a huge burden on the system.

The problem of the 'association' approach is that then every AID will have to periodically scan the whole table to re-balance the load for a use case like an APIC could go down then come back up again. While in the 'owndership" solution, only the extra AID will need to be in the recovery mode.

For the duplication code concern, I've consolidated those so there is no more duplication code now. Thx.

aim/agent/aid/universes/aci/aci_universe.py
rkukura: OK, I missed where the exception was being handled in _update_apic_assign_db(). I see that None will be returned and the loop will try the next URL.

PR #266, opened 2019-10-07
Override default for use_stderr to False

skandasw:
In older releases of oslo_log, the use_stderr cfg_opt has a default
of True. This results in logs getting replicated to /var/log/messages
in addition to the targetted log file as seen in newton deployments.
Being unnecessary duplication the community has fixed this in later
releases, we set it to False in the aim.conf template for installers
to pick up.


tbachman:
No longer needed, since we aren't supporting newton.

PR #267, opened 2019-10-18
Use annotation instead of Tags for ownership

kentwu1:
Starting F, APIC introduced the concept of annotation, designed
to let orchestrators declare ownership over managed objects.
The APIC team wants us to use annotation instead of tags because
the latter impact performance greately, both in terms of memory
and CPU cycle (a task is generated for every new tag).

This patch want to keep compatibility with APIC back and forth,
so we first test APIC for annotation support and then start using
it for ownership, falling back to tags if support is missing.
On upgrade, I expect AIM to just suddenly forget about the
existing tags and push annotations around. This might require an
external script to get rid of the hanging tags.

### DIFF
--- a/aim/agent/aid/universes/aci/tenant.py
+++ b/aim/agent/aid/universes/aci/tenant.py
@@ -44,6 +44,7 @@ CHILDREN_LIST = set(converter.resource_map.keys() + ['fvTenant', 'tagInst'])
 TOPOLOGY_CHILDREN_LIST = ['fabricPod', 'opflexODev', 'fabricTopology']
 CHILDREN_MOS_UNI = None
 CHILDREN_MOS_TOPOLOGY = None
+SUPPORTS_ANNOTATIONS = None
 RESET_INTERVAL = 3600
 DEFAULT_WS_TO = '900'
 
@@ -94,6 +95,14 @@ def get_children_mos(apic_session, root):
                     raise e
                 CHILDREN_MOS_TOPOLOGY.add(mo_name)
         return CHILDREN_MOS_TOPOLOGY
+
+
+def supports_annotations(apic_session):
+    global SUPPORTS_ANNOTATIONS
+    if SUPPORTS_ANNOTATIONS is None:
+        tn = apic_session.GET('/mo/uni/tn-common.json')
+        SUPPORTS_ANNOTATIONS = 'annotation' in tn[0]['fvTenant']['attributes']
+    return SUPPORTS_ANNOTATIONS
 
 
 OPERATIONAL_LIST = [FAULT_KEY]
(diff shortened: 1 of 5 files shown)


aim/agent/aid/universes/aci/tenant.py
```diff
@@ -436,41 +595,29 @@ def _push_aim_resources(self):
...
-                                                        **attr)
-                                # Object creation was successful, change object
-                                # state
+                                LOG.debug("UPDATING in APIC: %s" % to_update)
+                                # Update object ownership
+                                self._post_with_transaction(to_update,
+                                                            modified=True)
+                            # Object creation was successful, change object
+                            # state
+                            if method == base_universe.CREATE:
```
kentwu1: this line is redundant of line#599. So we should move self.creation_succeeded(aim_object) up to around line#605.

tbachman:
Just pushed an update, rebasing and addressing reviewer comments.

tbachman:
This still needs a fix - for an upgrade, it should create annotations for old resources and remove tags from them. I did a functional test in the lab and saw that annotations were used for new resources, but the old ones weren't replaced.

kentwu1:
From Ivar's commit msg,

"On upgrade, I expect AIM to just suddenly forget about the
existing tags and push annotations around. This might require an
external script to get rid of the hanging tags."

it should start to use annotation even for old resources after an upgrade. Is this not happening? He did mention that we will probably need an external script to get rid of those hanging tags though.

tbachman:
There are also some parts of the code that still look at "tag-". We probably need to move those to the new manager class, in order to handle both cases (tags or annotations).

aim/agent/aid/universes/aci/tenant.py
```diff
@@ -158,6 +167,139 @@ def _get_instance_subscription_urls(self,
...
+                    dn = obj.values()[0]['attributes']['dn']
+                    dn += '/tag-%s' % self.tag_key
+                    tags.append({"%s__%s" % (TAG_KEY, obj.keys()[0]):
+                                 {"attributes": {"dn": dn}}})
+            result.extend(tags)
+        return result
+
+    def set_ownership_change(self, to_push):
+        to_update = []
+        to_delete = []
```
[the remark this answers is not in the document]
tbachman: Oops - I was trying to fix it to be like the original patch and forgot to change the to_push returned back to to_delete. Will update with that fix.

aim/agent/aid/universes/aci/tenant.py
```diff
@@ -436,42 +594,27 @@ def _push_aim_resources(self):
...
-                                        getattr(getattr(self.aci_session, mo),
-                                                method)(*rns, transaction=trs,
-                                                        **attr)
-                                # Object creation was successful, change object
-                                # state
-                                self.creation_succeeded(aim_object)
+                                LOG.debug("UPDATING in APIC: %s" % to_update)
+                                # Update object ownership
+                                self._post_with_transaction(to_update,
+                                                            modified=True)
```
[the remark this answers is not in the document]
tbachman: Yes, I think so - will fix.

tbachman:
Just pushed an update, addressing reviewer comments.

aim/agent/aid/universes/aci/tenant.py
```diff
@@ -158,6 +167,139 @@ def _get_instance_subscription_urls(self,
...
+                    if dn.endswith('/tag-%s' % self.tag_key):
+                        dec = apic_client.DNManager().aci_decompose_dn_guess(
+                            dn, obj.keys()[0])
+                        parent_dec = dec[1][:-1]
+                        parent_dn = apic_client.DNManager().build(parent_dec)
+                        parent_type = parent_dec[-1][0]
+                        to_update.append(
+                            {parent_type: {'attributes': {'dn': parent_dn,
+                                                          'annotation': ''}}})
+                else:
```
kentwu1: I think we should get rid of this "else:" and always do "to_delete.append(obj)", otherwise those tags won't get deleted......

aim/agent/aid/universes/aci/tenant.py
```diff
@@ -436,42 +594,29 @@ def _push_aim_resources(self):
                                 aim_object.pre_existing = True
                             to_push = self.to_aci_converter.convert(
                                 [aim_object])
```
kentwu1: I see this debug log "LOG.debug('%s AIM object %s in APIC' % (method, repr(aim_object)))" was removed. Can we put it back then as I don't see that info is being kept somewhere else so.

aim/agent/aid/universes/aci/tenant.py
tbachman: Good point - will remove. And in fact this goes back to your earlier recommendation to get rid of to_delete all-together - it's the same as to_push, so I'll just use tat.

aim/agent/aid/universes/aci/tenant.py
tbachman: Will put back.

tbachman:
Pushed update addressing reviewer comments.

kentwu1:
LGTM now.

PR #271, opened 2019-11-08
Fix deadlocks

tbachman:
Detect and resolve any deadlock condition in the main aim-aid
deamon loop.


skandasw:
Looks good to me.

PR #272, opened 2019-12-03
Fix UT defaults

tbachman:
The UT framework supports checking against a set of default
values when a resource in AIM is created using just the
identity fields. Commit 983d5e677949454f3880049e4587f3c49bd53349
broke this feature, so this patch re-introduces it.


tbachman:
Please don't merge this yet - it should go in after we've completed the release.

tbachman:
Actually,  was confused - it was broken with that commit. The commit moved the initialization from the class to the setup. Although the class is included via MI, when the test is run, the setUp method is called, which clears this dict.

PR #273, opened 2019-12-09
Add mode for EPG static ports

tbachman:
The current code for supporting static port configuration on EPGs
lacks the ability to configure the mode. ACI offers 3 modes:

1) "regular". This is a trunked configuration
2) "native". This is an 802.1p configuration
3) "untagged". This is an untagged configuration with a native VLAN

This patch adds the ability to configure the mode setting, and uses
a default setting of "regular".  If the mode setting isn't provided,
the default value is populated in the returned resource, in order
to maintain backwards-compatibility. Updates that exclude the mode,
provided it currently has the default value of "regular", will be
considered no-ops if no other attributes are changed.

The patch includes a DB migration, which adds this column to the table,
and populates rows in the table with the default value ("regular"), since
this was the default value for any such configuration in AIM.


tbachman:
Please do not merge yet - this should go in after the release

irathore:
+1

aim/aim_manager.py
```diff
@@ -203,7 +203,7 @@ def update(self, context, resource, fix_ownership=False,
                     if old_resource.user_equal(
```
kentwu1: It seems to me that old_resource_copy will have all the fields available including this new field 'mode' as it was built from DB in line# 191. If that's the case then we don't really need the new API "def attributes_equivalent(superset, subset)" defined in resource.py.

aim/api/resource.py
```diff
@@ -109,6 +109,32 @@ def make_serializable(key, attr):
...
+        #          types, then we can just use equivalency directly. If
+        #          they are non-scalar (e.g. dictionaries), then we just
+        #          check if the new one is a subset of the existing one.
+        #          This was added because there are default values present
+        #          in the model representation, and if an update leaves
+        #          these fields out, then we don't want the equivalency
+        #          check to fail. This solution will not work if there is
+        #          a resource that uses non-scalar types in the list
+        #          that are allowed to have some elements absent.
+        def attributes_equivalent(superset, subset):
```
kentwu1: Per my previous comment, I don't think this is really needed.

aim/api/types.py
```diff
@@ -41,6 +41,7 @@
                '88': 'eigrp', '89': 'ospfigp', '103': 'pim', '115': 'l2tp'}
 
 ethertype = {'0': UNSPECIFIED, '1': 'ipv4', '2': 'ipv6'}
+spmodes = {'0': 'regular', '1': 'native', '2': 'untagged'}
```
kentwu1: In ACI on-line help, those 3 modes are Trunk, Access & Access Untagged. Should we just follow the naming used there instead then?
https://172.28.184.50/help/content/index.html#fv_infoAEPgStaticBindings.html

aim/aim_manager.py
tbachman: I'm not sure I follow. The new code checks if it's a subset or not. Without it, an update request that leaves out the mode field would result in a new version of the object with an epoch bump, even though the objects are equivalent.

aim/api/resource.py
tbachman: see my comment above

aim/api/types.py
tbachman: I was looking at this:
https://172.28.184.30/doc/html/TYPE-fv-Mode.html

I wasn't sure which were the right types. The ones in the GUI are the ones you mentioned, but from an API perspective, I thought it was tricky having two that had the word "access", plus I wanted to keep the names short.

aim/api/types.py
```diff
@@ -142,6 +143,10 @@ def list_of_dicts(*dict_keys_and_types):
     "type": "object",
     "properties": {"encap": {"type": "string"},
                    "path": {"type": "string"},
+                   "mode": {"type": "string",
+                            "oneOf": [enum(*spmodes.values()),
+                                      {"type": "string",
```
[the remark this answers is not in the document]
tbachman: Good point. Will remove.

aim/db/models.py
```diff
@@ -376,7 +377,14 @@ def to_attr(self, session):
             static_path = {'path': p.path, 'encap': p.encap}
             if p.host:
                 static_path['host'] = p.host
+            static_path['mode'] = p.mode if p.mode else 'regular'
```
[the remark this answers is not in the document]
tbachman: Good point - I will change this to a hasattribute check.

aim/db/models.py
```diff
@@ -376,7 +377,14 @@ def to_attr(self, session):
...
             res_attr.setdefault('static_paths', []).append(static_path)
+        # If there is a static path set, but no mode provided,
+        # make sure that the default is populated
+        for p in res_attr.pop('static_paths', []):
```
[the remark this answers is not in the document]
tbachman: Good point - will see what I can do here.

aim/aim_manager.py
kentwu1: I think you meant if update_attr_val doesn't have the 'mode' attribute in it. But in line#202, you can see that the code applies those update to the old_resource variable which was built in line#191 from db_obj which is from DB. So old_resource should already have that 'mode' field even though the update_attr_val doesn't have it. Make sense?

aim/api/types.py
kentwu1: I'm fine with it if that's what was defined in the API level.

aim/aim_manager.py
tbachman: The problem is that the setattr is setting the entire static_path attribute using what was provided by the caller - not updating the individual dicts in the static_paths list.

kentwu1: Hmm......Wondering how would that work before if an update request leaving out the 'encap' field comes in?

tbachman: The model class won't create the associated entry in the DB, which means it won't get added to the list:
https://github.com/noironetworks/aci-integration-module/blob/master/aim/db/models.py#L342

That's the existing behavior (i.e. nothing changes with this patch).

kentwu1: Then I saw u added "mode=p.get('mode', 'regular')" in line#346 of models.py, that should be good enough right? It should be a similar problem if an update request leaving out the 'host' field comes in, how did we handle that before then?

tbachman: If host isn't set, it will set it to an empty string in the model class, but neither host or mode are requirements for creating a new entry in the DB (i.e. they don't appear in that if statement).

It's worth pointing out that this method suffers from the same issue that the security groups/rules suffered - namely that the list of DB entries are deleted and recreated each time (fixed in https://github.com/noironetworks/aci-integration-module/commit/3b1fd190293569f4e82902cf69322374e143a3cc#diff-27d21d9e0ec23b43a0e3f5d38ef04922). We probably should revisit all our objects that have lists and fix any that are doing the same thing.

aim/aim_manager.py
```diff
@@ -203,7 +203,7 @@ def update(self, context, resource, fix_ownership=False,
...
                             old_resource_copy) and not force_update:
                         # Nothing to do here
-                        return old_resource
+                        return old_resource_copy
```
kentwu1: nit: its probably the same thing but I feel that we should keep the existing code unchanged, i.e., just return the old_resource here like before.

PR #276, opened 2020-02-13
Add python-semantic_version dependency

ratnakark:
Added python-semantic_version package as a dependency


tbachman:
Thanks for the patch!

PR #278, opened 2020-02-25
Add the NestedParameter and VlanRange table

kentwu1:
These will be used by openShift 4.3 IPI install over openStack.


aim/db/infra_model.py
```diff
@@ -164,3 +165,54 @@ class ApicAssignment(model_base.Base, model_base.AttributeMixin):
...
                                       onupdate=func.now())
+
+
+class VlanRange(model_base.Base):
+    """DB model for vlan ranges under a NestedParameter."""
+
+    __tablename__ = 'aim_vlan_ranges'
+
+    nested_parameter_cluster_name = sa.Column(
+        sa.String(64), sa.ForeignKey('aim_nested_parameter.cluster_name'),
```
tbachman: Is 64 large enough to support the cluster name?

aim/api/infra.py
```diff
@@ -154,3 +155,43 @@ def is_available(self, context):
             LOG.debug("APIC %s is not available. Last update time was %s" %
                       (self.apic_host, self.last_update_timestamp))
             return False
+
+
+class NestedParameter(resource.ResourceBase):
```
tbachman: We probably should put a revisit that we need to deprecate these resources once there's a proper fix in the OpenShift Installer.

aim/db/infra_model.py
```diff
@@ -164,3 +165,54 @@ class ApicAssignment(model_base.Base, model_base.AttributeMixin):
     aim_aid_id = sa.Column(sa.String(64))
     last_update_timestamp = sa.Column(sa.TIMESTAMP, server_default=func.now(),
                                       onupdate=func.now())
+
+
+class VlanRange(model_base.Base):
```
tbachman: Same here - comment about deprecation.

aim/db/infra_model.py
```diff
@@ -164,3 +165,54 @@ class ApicAssignment(model_base.Base, model_base.AttributeMixin):
...
+    __tablename__ = 'aim_nested_parameter'
+
+    cluster_name = model_base.name_column(primary_key=True)
+    domain_name = model_base.name_column()
+    domain_type = sa.Column(sa.String(32), nullable=False)
+    domain_infra_vlan = sa.Column(sa.Integer, nullable=False)
+    domain_service_vlan = sa.Column(sa.Integer, nullable=False)
+    domain_node_vlan = sa.Column(sa.Integer, nullable=False)
+
+    vlan_range_list = orm.relationship(VlanRange,
```
tbachman: Whenever we have relationships that are lists, where the list elements aren't exposed as an AIM resource but are ORM objects, we need to make sure these are optimized or we run into performance problems due to deletion and recreation of DB entries b/c their ORM resources are deleted and recreated. We had to do a similar optimization to the SecurityGroupRuleRemoteIp's in this patch: https://github.com/noironetworks/aci-integration-module/commit/3b1fd190293569f4e82902cf69322374e143a3cc

aim/db/migration/alembic_migrations/versions/8c6cbd136805_nested_parameter.py
```diff
@@ -0,0 +1,59 @@
...
+        'aim_nested_parameter',
+        sa.Column('cluster_name', sa.String(64), nullable=False),
+        sa.Column('domain_name', sa.String(64)),
+        sa.Column('domain_type', sa.String(32), nullable=False),
+        sa.Column('domain_infra_vlan', sa.Integer, nullable=False),
+        sa.Column('domain_service_vlan', sa.Integer, nullable=False),
+        sa.Column('domain_node_vlan', sa.Integer, nullable=False),
+        sa.Column('epoch', sa.BigInteger(), nullable=False,
+                  server_default='0'),
+        sa.PrimaryKeyConstraint('cluster_name'))
```
tbachman: Can cluster names overlap across OpenStack projects (and therefore AIM tenants)? If so, we may want to add tenant name in here as a primary key, or create a UUID to use to make sure this is unique.

aim/tests/unit/test_aim_manager.py
```diff
@@ -242,13 +242,19 @@ def _test_resource_ops(self, resource, test_identity_attributes,
         # Verify successful creation
         r1 = self.mgr.create(self.ctx, res)
         for k, v in creation_attributes.iteritems():
-            self.assertEqual(v, getattr_canonical(r1, k))
+            if isinstance(v, list):
```
tbachman: Are you sure this is needed? We have other objects that use lists (e.g. SecurityGroupRule, EndpointGroup)

aim/api/infra.py
kentwu1: sure will do.

aim/db/infra_model.py
kentwu1: will do.

aim/db/infra_model.py
kentwu1: aim_nested_parameter.cluster_name is also a string of length 64 so it should be good.

aim/db/infra_model.py
kentwu1: Since this nested parameter thing is only for user configuration and is not part of any openStack or ACI resources so this might be ok for now? Especially we might have to back this out once they got the installer change working.

aim/tests/unit/test_aim_manager.py
kentwu1: In my testing, it also works sometimes if I change the elements of that list so I guess we were probably just lucky before. I think this is also because all our UTs probably don't have a long list of SGRs or EPGs to test with. But this change does make sense to me though because 2 lists should be considered equal as long as their elements are all the same even if the order is different.

aim/db/migration/alembic_migrations/versions/8c6cbd136805_nested_parameter.py
kentwu1: Good point. Will add a project_id column and make it the primary key also along with the cluster name.

kentwu1:
pushed a revision to address those review comments.

kentwu1:
pushed another revision to use tenant_id instead of tenant_name.

aim/api/infra.py
```diff
@@ -154,3 +155,46 @@ def is_available(self, context):
...
             return False
+
+
+# REVISIT(kentwu): We will need to deprecate this once there is
+# a proper fix in the openShift IPI installer.
+class NestedParameter(resource.ResourceBase):
+    """Nested parameters needed for openShift on openStack install"""
+
+    identity_attributes = t.identity(
+        ('tenant_id', t.name),
```
rkukura: What is tenant_id here? I notice this is not tenant_name, which is generally an ACI tenant name. If this is supposed to be an OpenStack project_id, which I suspect based on the GBP repo patch (similar comment coming there), I'd strongly prefer you call it project_id rather than tenant_id. The term tenant has been deprecated in OpenStack for years, and in AIM, project_id is a lot less likely to be confused with tenant_name.

kentwu1: ok will change it to project_id then.

kentwu1:
pushed a revision to address the latest review comment.

PR #281, opened 2020-04-30
Tag handling commands

tbachman:
Add AIM CLI commands for finding and deleting tags in ACI.

### DIFF
--- a/aim/tests/unit/tools/cli/test_manager.py
+++ b/aim/tests/unit/tools/cli/test_manager.py
@@ -17,6 +17,7 @@ import ast
 import mock
 import six
 
+
 from aim import aim_manager
 from aim.api import infra
 from aim.api import resource
@@ -388,6 +389,49 @@ class TestManager(base.TestShell):
             self.run_command('manager sync-state-recover')
             # Items are updated
             self.assertEqual(4, up.call_count)
+
+    def test_tag_list_command(self):
+        apic_manager_object = mock.Mock()
+
+        def get_mgr():
+            return apic_manager_object
+
+        mo_dict = {'tagInst': {'attributes':
+                   {'dn': 'uni/tn-tn1/tag-openstack_aid'}}}
+        apic_manager_object.apic.list_mo = mock.Mock(
+            return_value=[mo_dict])
+        apic_manager_object.apic.transaction = mock.MagicMock()
+
+        with mock.patch('aim.tools.cli.commands.infra.get_apic_manager',
+                        side_effect=get_mgr):
+            with mock.patch('click.echo') as ce:
+                self.run_command('infra tag-list -t openstack_aid')
+        ce.called_once_with('uni/tn-tn1/tag-openstack_aid')
+
+    def test_tag_delete_command(self):
+        apic_manager_object = mock.Mock()
+
+        def get_mgr():
+            return apic_manager_object
+        apic_manager_object.apic.DELETE = mock.Mock()
+        apic_manager_object.apic.transaction = mock.MagicMock()
+        # First try and delete the tenant, which should fail
+        with mock.patch('aim.tools.cli.commands.infra.get_apic_manager',
+                        side_effect=get_mgr):
+            with mock.patch('click.echo') as ce:
+                self.run_command('infra tag-delete --dn uni/tn-tn1')
+        ce.called_once_with("uni/tn-tn1 object isn't a tag -- ignoring")
+        apic_manager_object.apic.transaction.assert_not_called()
+
+        # Now actually delete a tag
+        with mock.patch('aim.tools.cli.commands.infra.get_apic_manager',
+                        side_effect=get_mgr):
+            with mock.patch('click.echo') as ce:
+                self.run_command(
+                    'infra tag-delete --dn uni/tn-tn1/tag-openstack_aid')
+        ce.called_once_with("uni/tn-tn1/tag-openstack_aid deleted")
+        apic_manager_object.apic.DELETE.assert_called_with(
+            '/mo/uni/tn-tn1/tag-openstack_aid.json')
 
 
 class TestManagerResourceOpsBase(object):
--- a/aim/tools/cli/commands/infra.py
+++ b/aim/tools/cli/commands/infra.py
@@ -56,3 +56,34 @@ def create(ctx):
     ctx.obj['apic_manager'].ensure_infra_created_on_apic()
     ctx.obj['apic_manager'].ensure_bgp_pod_policy_created_on_apic()
     ctx.obj['apic_manager'].ensure_opflex_client_cert_validation_disabled()
+
+
+@infra.command(name='tag-list')
+@click.option('--tag-name', '-t', help='Tag name')
+@click.pass_context
+def tag_list(ctx, tag_name):
+    """APIC command for listing tagged resources.
+
+    Lists all the resources in ACI with a tagInst.
+    """
+    with ctx.obj['apic_manager'].apic.transaction():
+        # Find all the MOs with the system ID
+        for mo in ctx.obj['apic_manager'].apic.list_mo(
+                ctx.obj['apic_manager'].apic.tagInst.mo, name=tag_name):
+            click.echo("%s" % mo['tagInst']['attributes']['dn'])
+
+
+@infra.command(name='tag-delete')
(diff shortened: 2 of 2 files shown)


aim/tools/cli/commands/infra.py
```diff
@@ -56,3 +56,32 @@ def create(ctx):
     ctx.obj['apic_manager'].ensure_infra_created_on_apic()
     ctx.obj['apic_manager'].ensure_bgp_pod_policy_created_on_apic()
     ctx.obj['apic_manager'].ensure_opflex_client_cert_validation_disabled()
+
+
+@infra.command(name='tag-list')
+@click.option('--tag-name', '-t', help='Tag name')
+@click.pass_context
+def tag_list(ctx, tag_name):
+    ctx.obj['apic_manager'].ensure_infra_created_on_apic()
```
kentwu1, line 65: Why do we have to call ensure_infra_created_on_apic() here?

aim/tools/cli/commands/infra.py
```diff
@@ -56,3 +56,32 @@ def create(ctx):
...
+        # Find all the MOs with the system ID
+        for mo in ctx.obj['apic_manager'].apic.list_mo(
+                ctx.obj['apic_manager'].apic.tagInst.mo, name=tag_name):
+            click.echo("%s" % mo['tagInst']['attributes']['dn'])
+
+
+@infra.command(name='tag-delete')
+@click.option('--dn', help='Distinguished Name (DN) for Managed Oobject (MO)')
+@click.pass_context
+def tag_delete(ctx, dn, **kwargs):
```
kentwu1, line 80: Looks like this API can pretty much delete all the MOs, not just the tagged ones. Should we name it differently or at least change the comments accordingly then?

aim/tools/cli/commands/infra.py
```diff
@@ -56,3 +56,32 @@ def create(ctx):
     ctx.obj['apic_manager'].ensure_infra_created_on_apic()
     ctx.obj['apic_manager'].ensure_bgp_pod_policy_created_on_apic()
     ctx.obj['apic_manager'].ensure_opflex_client_cert_validation_disabled()
+
+
+@infra.command(name='tag-list')
```
kentwu1, line 61: Is that possible to add UTs for these 2 APIs?

aim/tools/cli/commands/infra.py
tbachman: Oops. We don't. Will remove.

aim/tools/cli/commands/infra.py
tbachman: Good point - I should add a guard to make sure it's for a tag-inst. We don't want to delete any others I think, as that should just be done using aim resources (there are no resources for tags, which is why we added this command).

aim/tools/cli/commands/infra.py
tbachman: Will look into that.

tbachman:
Pushed an update, addressing reviewer comments.

aim/tools/cli/commands/infra.py
```diff
@@ -56,3 +56,34 @@ def create(ctx):
...
+@click.pass_context
+def tag_list(ctx, tag_name):
+    """APIC command for listing tagged resources.
+
+    Lists all the resources in ACI with a tagInst.
+    """
+    with ctx.obj['apic_manager'].apic.transaction():
+        # Find all the MOs with the system ID
+        for mo in ctx.obj['apic_manager'].apic.list_mo(
+                ctx.obj['apic_manager'].apic.tagInst.mo, name=tag_name):
```
kentwu1, line 72: wondering what's the magic here so that we can access the tagInst MO directly thru "ctx.obj['apic_manager'].apic.tagInst.mo".....I couldn't find the magic in the apic_client.py file under apicapi repo.

aim/tests/unit/tools/cli/test_manager.py
```diff
@@ -389,6 +390,52 @@ def test_sync_state_recover(self):
             # Items are updated
             self.assertEqual(4, up.call_count)
 
+    def test_tag_list_command(self):
+        self.mgr.create(self.ctx, resource.Tenant(name='tn1'))
+        mo_dict = {'tagInst': {'attributes':
+                   {'dn': 'uni/tn-tn1/tag-openstack_aid'}}}
+        mock_mo = mock.MagicMock()
+        mock_mo.__getitem__.side_effect = mo_dict.__getitem__
+        MO_LIST = [mock_mo]
```
kentwu1, line 399: We can't just do "MO_LIST = [mo_dict]" here?

aim/tests/unit/tools/cli/test_manager.py
```diff
@@ -389,6 +390,52 @@ def test_sync_state_recover(self):
...
+        apic_manager_object.apic.list_mo = mock.Mock(
+            return_value=MO_LIST)
+        apic_manager_object.apic.transaction = mock.MagicMock()
+        with mock.patch('aim.tools.cli.commands.infra.get_apic_manager',
+                        side_effect=get_mgr):
+            self.run_command('infra tag-list -t openstack_aid')
+        # REVISIT: Find a better way of creating the verify call object.
+        # I did this because I couldn't find another way of creating an
+        # equivalent call object
+        verify_mo = mock.MagicMock()
```
kentwu1, line 413: Maybe we can just create a mock for click.echo then verify that its getting called once with the parameter "uni/tn-tn1/tag-openstack_aid"?

aim/tools/cli/commands/infra.py
tbachman: It's using the __getattr__:
https://github.com/noironetworks/apicapi/blob/master/apicapi/apic_client.py#L1074-L1079

aim/tests/unit/tools/cli/test_manager.py
tbachman: If I change it to use the mock on click.echo, then yes we could. I did it on the mock_mo so that I could verify that calls.

aim/tests/unit/tools/cli/test_manager.py
tbachman: Will look into that.

tbachman:
Pushed update, based on reviewer coments.

aim/tests/unit/tools/cli/test_manager.py
```diff
@@ -389,6 +390,51 @@ def test_sync_state_recover(self):
...
+        def get_mgr():
+            return apic_manager_object
+
+        mo_dict = {'tagInst': {'attributes':
+                   {'dn': 'uni/tn-tn1/tag-openstack_aid'}}}
+        apic_manager_object.apic.list_mo = mock.Mock(
+            return_value=[mo_dict])
+        apic_manager_object.apic.transaction = mock.MagicMock()
+
+        self.mgr.create(self.ctx, resource.Tenant(name='tn1'))
```
kentwu1, line 405: nit: do we really need to create this tenant for this UT?

aim/tests/unit/tools/cli/test_manager.py
```diff
@@ -389,6 +390,51 @@ def test_sync_state_recover(self):
...
+        with mock.patch('aim.tools.cli.commands.infra.get_apic_manager',
+                        side_effect=get_mgr):
+            with mock.patch('click.echo') as ce:
+                self.run_command('infra tag-list -t openstack_aid')
+        ce.called_once_with('uni/tn-tn1/tag-openstack_aid')
+
+    def test_tag_delete_command(self):
+        self.mgr.create(self.ctx, resource.Tenant(name='tn1'))
```
kentwu1, line 413: ditto

aim/tools/cli/commands/infra.py
kentwu1: Ack.

aim/tests/unit/tools/cli/test_manager.py
tbachman: Yeah - I can remove it. My initial thinking was that we would somehow return that object, but then just decided to mock it. Will remove.

aim/tests/unit/tools/cli/test_manager.py
tbachman: ditto ;)

tbachman:
Pushed update, based on reviewer comments.

PR #282, opened 2020-05-06
Add the remote_group_id column to the SG_rule table

kentwu1:
1. Also prevent the SG rule MO from showing up in APIC by
removing it from the AIM tree when its a remote_group_rule
and its remote_ips list is empty.

2. Also remove the temporary config parameter fetch_sgr_from_db
because we have been using it for a while without issues + we
will need to do that always now for item #1 to work.

3. Also return the hash tree result properly in the
_get_hash_trees() API otherwise the test_tree_result() is a
no-op.


aim/tests/unit/test_aim_manager.py
```diff
@@ -1356,11 +1360,12 @@ class TestSecurityGroupRuleMixin(object):
                                 'security_group_subject_name': 'subject1',
                                 'name': 'rule1',
                                 'direction': 'ingress',
-                                'remote_ips': ['10.0.0.1/30',
```
tbachman, line 1359: By removing these remote IPs, do we lose some coverage for a UT?

kentwu1: It shouldn't because I added those back in the test_update_attributes.

kentwu1:
pushed a revision to address the review comments.

aim/db/migration/alembic_migrations/versions/f0c056954eee_sg_rule_remote_group_id.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2017 Cisco, Inc.
```
rkukura, line 1: Its 2020!

PR #286, opened 2020-05-26
Make the deadlock time configurable

kentwu1:


aim/agent/aid/service.py
```diff
@@ -305,6 +306,10 @@ def _change_squash_time(self, new_conf):
         # TODO(ivar): interrupt current sleep and restart with new value
         self.squash_time = new_conf['value']
 
+    def _change_deadlock_time(self, new_conf):
+        # TODO(ivar): interrupt current sleep and restart with new value
```
tbachman, line 310: Can you change this from TODO(ivar) to REVISIT? Doesn't seem fair to list this as a todo for Ivar any more ;-).

PR #287, opened 2020-05-28
Skip those duplicate SG rules

kentwu1:
We still need to make sure those skipped action logs
got deleted at the end though. This patch will improve
the action logs process time considerably especially
for those SG rules with huge remote_ips list.

### DIFF
--- a/aim/db/hashtree_db_listener.py
+++ b/aim/db/hashtree_db_listener.py
@@ -202,6 +202,7 @@ class HashTreeDbListener(object):
         log_by_root = {}
         resource_paths = ('resource', 'service_graph', 'infra', 'tree',
                           'status')
+        sg_rule_logs = {}
         for log in logs:
             if log.action == aim_tree.ActionLog.RESET:
                 resetting_roots.add(log.root_rn)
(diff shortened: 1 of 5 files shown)


aim/db/hashtree_db_listener.py
```diff
@@ -253,8 +255,29 @@ def _preprocess_logs(self, ctx, logs):
...
             log_by_root.setdefault(log.root_rn, []).append(
                 (action, aim_res, log))
+
+        for sg_rule_name, list_of_logs in sg_rule_logs.items():
+            # We will skip the processing of all the sg_rules except
+            # for the last one.
+            for index, items in enumerate(list_of_logs[:-1]):
+                items[0] = aim_tree.ActionLog.SKIP
+                list_of_logs[index] = tuple(items)
+            # Convert the last one also.
```
tbachman, line 276: The intent of lines 240-253 were to just use what's actually in the DB, rather than iterate through all the actions, with the idea being that the DB (or at least the DB within this transaction) has the desired end state. Should the loop instead just be looking for the last entry with the matching state (I don't know how/why it won't be the last one, but it might be worth checking)?

kentwu1: The last one could also be a DELETE operation though. Its true that we will only need the last one however we still need to pass the whole list to the log_by_root dictionary because that's the only way we can guarantee those action logs got deleted later on here,
https://github.com/noironetworks/aci-integration-module/blob/master/aim/db/hashtree_db_listener.py#L319

But I think I don't have to create another loop here at line#273 because I can mark those entries as SKIP while queuing up those SG rules related action logs around line#264. Let me do that then, which can speed things up.

PR #288, opened 2020-05-30
Scope the catch_up_with_action_logs() by each tenant

kentwu1:
Inside this API, it will call _push_changes_to_trees() which will
then call _delete_logs() for each tenant respectively. What we
found from customer debug logs shows that if aim-aid got killed
by the watchdog while in the middle of _push_changes_to_trees()
due to some spike of action logs, those deleted action logs will
show up again even in the next restart because they were rolled
back by the outmost transaction. By scoping it with each tenant,
the backlog of action logs should reduce after each restart, which
will allow aim-aid to gradually recover itself instead of keep
getting killed by the watchdog. This change could potentially
increase the number of 'serve' and 'reconcile' events posted to
the event queue consumed by aim-aid. However since there is already
a sqash mechanism in place when aim-aid is picking up those events,
this should be ok too.

Also moved the test_tree_hooks_transactions() into a class
without using the mock store. This way it would operate just like
in the real world setup. Otherwise it will automatically call
catch_up_with_action_logs() by end of each transaction and this
effect can be nested again and again too, which makes it very tricky
to really see what we want to test about.

### DIFF
--- a/aim/db/hashtree_db_listener.py
+++ b/aim/db/hashtree_db_listener.py
@@ -178,24 +178,31 @@ class HashTreeDbListener(object):
         return cache[klass]
 
     def catch_up_with_action_log(self, store, served_tenants=None):
+        served_tenants = served_tenants or set()
         ctx = utils.FakeContext(store=store)
-        with ctx.store.begin(subtransactions=True):
-            served_tenants = served_tenants or set()
-            to_init = set(self.tt_mgr.retrieve_uninitialized_roots(ctx))
-            served_tenants |= to_init
-            # Nothing will happen if there's no action log
-            kwargs = {'order_by': ['root_rn', 'id']}
-            if served_tenants:
-                kwargs['in_'] = {'root_rn': served_tenants}
-            logs = self.aim_manager.find(ctx, aim_tree.ActionLog, **kwargs)
-            LOG.debug('Processing action logs: %s' % logs)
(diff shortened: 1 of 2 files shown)


aim/db/hashtree_db_listener.py
```diff
@@ -179,23 +179,22 @@ def _retrieve_class_root_type(self, klass, cache=None):
...
-            logs = self.aim_manager.find(ctx, aim_tree.ActionLog, **kwargs)
-            LOG.debug('Processing action logs: %s' % logs)
-            log_by_root, resetting_roots = self._preprocess_logs(ctx, logs)
-            self._cleanup_resetting_roots(ctx, log_by_root, resetting_roots)
-            self._push_changes_to_trees(ctx, log_by_root)
-            # REVISIT: This is temporary code for verifying solutions
-            # to concurrency issues. Remove when no longer needed.
-            if aim_cfg.CONF.aim.validate_config_trees:
-                self._validate_config_trees(ctx, log_by_root.keys())
+        served_tenants = served_tenants or set()
```
tbachman, line 182: I think we want to keep an outer transaction, but we should instead scope it by each tenant. This has the downside of more DB round trips (one per-tenant, instead of one for all tenants), but ensures that any operations for a tenant related to action logs can still be rolled-back (and therefore retried) if there's an exception.

kentwu1:
Pushed a revision to address the review comments.

PR #295, opened 2020-08-17
Code to enable netflow functionality in openstack setups

tanyatukade:
- Code to enable netflow functionality on opnestack setups.
- Added code to establish relationship to vmm domain.
- Tested on live fabric to make sure policies are pushed to leaf and compute nodes.

### DIFF
--- a/aim/api/resource.py
+++ b/aim/api/resource.py
@@ -1450,3 +1450,49 @@ class QosDppPol(AciResourceBase):
                                          'admin_st': 'enabled',
                                          'sharing_mode': 'dedicated',
                                          'monitored': False}, **kwargs)
+
+
+class VmmVswitchPolicyGroup(AciResourceBase):
+    """Resource representing VSwitch Policy Group in ACI.
+
+    Identity attributes are domain_type, domain_name.
+    """
+
+    identity_attributes = t.identity(
+        ('domain_type', t.name),
+        ('domain_name', t.name))
+    other_attributes = t.other(
+        ('monitored', t.bool),
+        ('display_name', t.name))
+
+    _aci_mo_name = 'vmmVSwitchPolicyCont'
+    _tree_parent = VMMDomain
+
+    def __init__(self, **kwargs):
+        super(VmmVswitchPolicyGroup, self).__init__({'monitored': False},
+                                                    **kwargs)
+
+
+class VmmRelationToExporterPol(AciResourceBase):
+    """Resource representing Relationship to VMM Netflow Exporter Policy.
+
+    Identity attributes are domain_type, domain_name and tDn.
+    """
+
+    identity_attributes = t.identity(
+        ('domain_type', t.name),
+        ('domain_name', t.name),
+        ('netflow_path', t.string()))
+    other_attributes = t.other(
+        ('monitored', t.bool),
+        ('active_flow_time_out', t.integer),
+        ('idle_flow_time_out', t.integer),
+        ('sampling_rate', t.integer))
+
+    _aci_mo_name = 'vmmRsVswitchExporterPol'
+    _tree_parent = VmmVswitchPolicyGroup
+
+    def __init__(self, **kwargs):
+        super(VmmRelationToExporterPol, self).__init__(
+            {'monitored': False, 'active_flow_time_out': 60,
+             'sampling_rate': 0, 'idle_flow_time_out': 15}, **kwargs)
--- a/aim/db/migration/alembic_migrations/versions/4caed435b0cd_add_netflow.py
+++ b/aim/db/migration/alembic_migrations/versions/4caed435b0cd_add_netflow.py
@@ -29,6 +29,7 @@ depends_on = None
 
 from alembic import op
 import sqlalchemy as sa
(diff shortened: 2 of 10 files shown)


aim/db/migration/alembic_migrations/versions/4caed435b0cd_add_netflow.py
```diff
@@ -58,6 +59,47 @@ def upgrade():
...
+                  server_default='0'),
+        sa.PrimaryKeyConstraint('aim_id'),
+        sa.UniqueConstraint('domain_type', 'domain_name', 'netflow_path',
+                            name='uniq_aim_reln_exporter_identity'),
+        sa.Index('idx_aim_reln_exporter_identity',
+                 'domain_type', 'domain_name', 'netflow_path'),
+        sa.ForeignKeyConstraint(
+            ['domain_type', 'domain_name'],
+            ['aim_vmm_domains.type', 'aim_vmm_domains.name'],
+            name='fk_reln'))
```
kentwu1, line 101: I will suggest to spell out the full name like fk_vmm_reln_exporter_pol instead of just use 'fk_reln' here to minimize the chance of duplicate names.

aim/db/migration/alembic_migrations/versions/4caed435b0cd_add_netflow.py
```diff
@@ -58,6 +59,47 @@ def upgrade():
...
+                  server_default='0'),
+        sa.PrimaryKeyConstraint('aim_id'),
+        sa.UniqueConstraint('domain_type', 'domain_name',
+                            name='uniq_aim_vswitch_pol_grp_identity'),
+        sa.Index('idx_aim_vswitch_pol_grp_identity',
+                 'domain_type', 'domain_name'),
+        sa.ForeignKeyConstraint(
+            ['domain_type', 'domain_name'],
+            ['aim_vmm_domains.type', 'aim_vmm_domains.name'],
+            name='fk_vswitch_pol_grp'))
```
kentwu1, line 79: its better to spell out the full name like fk_vmm_vswitch_pol_grp here.

aim/db/migration/alembic_migrations/versions/4caed435b0cd_add_netflow.py
```diff
@@ -58,6 +59,47 @@ def upgrade():
...
+        sa.Column('display_name', sa.String(256), nullable=False, default=''),
+        sa.Column('monitored', sa.Boolean, nullable=False, default=False),
+        sa.Column('epoch', sa.BigInteger(), nullable=False,
+                  server_default='0'),
+        sa.PrimaryKeyConstraint('aim_id'),
+        sa.UniqueConstraint('domain_type', 'domain_name',
+                            name='uniq_aim_vswitch_pol_grp_identity'),
+        sa.Index('idx_aim_vswitch_pol_grp_identity',
+                 'domain_type', 'domain_name'),
+        sa.ForeignKeyConstraint(
```
tbachman, line 76: We actually don't want to have foreign keys here - an AIM design principle allows for configuration of a resource in the DB even if the parent object hasn't yet been configured (and this reminds me that we need to fix the QoS patch, as it also introduced ForeignKey relationships which aren't needed). The exception is when a property maps to a DB table but not an AIM resource (see SecurityGroupRuleRemoteIp).

aim/db/migration/alembic_migrations/versions/4caed435b0cd_add_netflow.py
```diff
@@ -58,6 +59,47 @@ def upgrade():
...
+        sa.Column('active_flow_time_out', sa.Integer),
+        sa.Column('idle_flow_time_out', sa.Integer),
+        sa.Column('epoch', sa.BigInteger(), nullable=False,
+                  server_default='0'),
+        sa.PrimaryKeyConstraint('aim_id'),
+        sa.UniqueConstraint('domain_type', 'domain_name', 'netflow_path',
+                            name='uniq_aim_reln_exporter_identity'),
+        sa.Index('idx_aim_reln_exporter_identity',
+                 'domain_type', 'domain_name', 'netflow_path'),
+        sa.ForeignKeyConstraint(
```
tbachman, line 98: Same here

aim/api/resource.py
```diff
@@ -1450,3 +1450,48 @@ def __init__(self, **kwargs):
...
+    """
+
+    identity_attributes = t.identity(
+        ('domain_type', t.name),
+        ('domain_name', t.name),
+        ('netflow_path', t.string()))
+    other_attributes = t.other(
+        ('monitored', t.bool),
+        ('active_flow_time_out', t.integer),
+        ('idle_flow_time_out', t.integer))
```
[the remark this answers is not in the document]
tanyatukade: Added in the new commit

aim/db/migration/alembic_migrations/versions/4caed435b0cd_add_netflow.py
tanyatukade: That makes sense. I have made changes in the new commit.

PR #297, opened 2020-08-17
Add debug to action log processing

tbachman:
When action logs are backed up, aim-aid logging should indicate
how many action logs are being processed. This commit adds a log
message at the INFO level if the number of action log messages to
be processed for a given tenant goes above 1,000.

### DIFF
--- a/aim/db/hashtree_db_listener.py
+++ b/aim/db/hashtree_db_listener.py
@@ -28,6 +28,7 @@ from aim.common import utils
 from aim import config as aim_cfg
 from aim import tree_manager
 
+ACTION_LOG_THRESHOLD = 1000
 MAX_EVENTS_PER_ROOT = 10000
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


aim/db/hashtree_db_listener.py
```diff
@@ -28,6 +28,7 @@
 from aim import config as aim_cfg
 from aim import tree_manager
 
+ACTION_LOG_THRESHOLD = 1000
```
kentwu1, line 31: LGTM but just wondering should we make this threshold configurable just in case user complains either not seeing the logs (but the system is already slow) or seeing too many logs while system is still operational....

tbachman: If it were affecting a behavior I'd say yes - but in this case it's just a log message. The threshold number I picked is a bit arbitrary, but in a normal working setup it shouldn't get triggered.

PR #298, opened 2020-08-18
Fix AIM QoS DB

tbachman:
Commit 4ce65eda8f4779c840d1a213654151c774930c80 added support for
QoS features in ACI. The DB migrations included foreign key constriants
to other tables, which aren't needed for most AIM ORM objects, since
AIM allows child objects to be configured in the DB before their parent
(except in cases where the child ORM object isn't represented as an AIM
resource, such as SecurityGroupRuleRemoteIp). Changing DB migration
scripts is permitted, provided it happens before a release is created
that includes the DB migraiton.


satpatch:
Thanks for fixing it.

tbachman:
No problem - we should have caught this in our initial review, but we missed it. Thanks again for the patch!

PR #301, opened 2020-09-03
Adapting to the model change on apic where the relation type was changed to a named relation.

satpatch:


aim/agent/aid/universes/aci/converter.py
```diff
@@ -215,31 +215,27 @@ def qos_rs_ingress_dpp_pol(object_dict, otype, helper,
             return []
         for index, attr in enumerate(destination_identity_attributes):
             res_dict[attr] = id[index]
-        if object_dict['tDn'] != '':
-            master_id = apic_client.DNManager().aci_decompose_with_type(
-                object_dict['tDn'], 'qosDppPol')
-            res_dict['ingress_dpp_pol'] = master_id[1][1]
+        if 'tnQosDppPolName' in object_dict:
```
tbachman, line 218: Were any ACI images released with the previous model? Or was it only internal builds?

tbachman (approved):
LGTM, but had a question about the model change.

aim/agent/aid/universes/aci/converter.py
satpatch: internal builds only.

> Were any ACI images released with the previous model? Or was it only internal builds?

kentwu1 (approved):

PR #305, opened 2020-09-08
updated default values for netflow version and port

tanyatukade:


aim/db/models.py
```diff
@@ -101,7 +101,7 @@ class NetflowVMMExporterPol(model_base.Base, model_base.HasDisplayName,
     dst_addr = sa.Column(sa.String(64))
     dst_port = sa.Column(sa.String(16))
     src_addr = sa.Column(sa.String(64))
-    ver = sa.Column(sa.String(16))
+    ver = sa.Column(sa.Enum('v5', 'v9', 'cisco-v1'))
```
tbachman, line 104: This needs the corresponding change to the DB migrations. Unfortunately we've released the current AIM code with support for netflow, so you'll have to add a new DB migration for this instead of modify the existing one.

PR #307, opened 2020-10-05
Erspan vepg-summary

tanyatukade:
- dropped invalid and ver_enforced columns in erspan-vepg-summary table since we would never use it to set or read it.
- dropped route_ip, scope and ver these are admin implicit props and admin implicit props cannot be modified.
- Handle acc bundle group special case since it is a pre-existing and monitored object.


tanyatukade:
The field "vsg_aim_id" in table "aim_span_spanlbl" seems to be not used. I need to remove that too

tbachman:
Please squash the commits.

tanyatukade:
squashed commits. Please review

aim/db/migration/alembic_migrations/versions/6dbf5128c06e_drop_erspan_invalid_column.py
```diff
@@ -0,0 +1,46 @@
...
+revision = '6dbf5128c06e'
+down_revision = 'fccf2c4f6282'
+branch_labels = None
+depends_on = None
+
+from alembic import op
+
+
+def upgrade():
+    with op.batch_alter_table('aim_span_vepg_summary') as \
```
tbachman, line 32: Please use parentheses instead of backslashes.

aim/db/migration/alembic_migrations/versions/6dbf5128c06e_drop_erspan_invalid_column.py
```diff
@@ -0,0 +1,46 @@
...
+        batch_op:
+        batch_op.drop_column('invalid')
+        batch_op.drop_column('ver_enforced')
+        batch_op.drop_column('route_ip')
+        batch_op.drop_column('scope')
+        batch_op.drop_column('ver')
+
+    with op.batch_alter_table('aim_span_spanlbl') as \
```
tbachman, line 40: Same comment here.

aim/db/migration/alembic_migrations/versions/6dbf5128c06e_drop_erspan_invalid_column.py
```diff
@@ -0,0 +1,46 @@
...
+        batch_op:
+        batch_op.drop_column('vsg_aim_id')
```
tbachman, line 42: I guess this means that we can just use the vsg_name field to reference the VSG?

tanyatukade: Thats right

aim/db/migration/alembic_migrations/versions/6dbf5128c06e_drop_erspan_invalid_column.py
tanyatukade: Removed blackslash

aim/agent/aid/universes/aci/converter.py
```diff
@@ -1149,6 +1153,22 @@ def bgp_as_id_converter(object_dict, otype, helper, to_aim=True):
                        'converter': fvRsCons_Ext_converter,
                        'convert_pre_existing': True,
                        'convert_monitored': False}],
+    'infraRsSpanVSrcGrp': [{'resource': resource.InfraAccBundleGroup,
```
tbachman, line 1156: Are these needed? Weren't they already added to the initial resource_map above?

tbachman (changes requested):
Just a question about possible duplication.

aim/agent/aid/universes/aci/converter.py
tanyatukade: You are right, that was duplicated. Removed it in the lastest commit.

tbachman (approved):

Issue #309, opened 2020-10-29
Scaling issue when updating static paths of an EndpointGroup


timeu:
The [codepath](https://github.com/noironetworks/aci-integration-module/blob/master/aim/db/models.py#L392) for updating the epg does not scale when having a high number of static paths (> 100). 
We are running into this issue in our production system when we try to scale up an existing 200 VM/hypervisor cluster by one VM because the port binding runs into a timeout (FYI: TAC case is open). 

We did some profiling on the live system of the [aim_manager.py](https://github.com/noironetworks/aci-integration-module/blob/master/aim/aim_manager.py#L190) codepath and most of the time is spent in the `context.store.from_attr` function (see [1] for line_profiler output).
Further investigation by enabling the SQLalchemy logging revealed that even when only one new static path is added a SELECT statement for all existing static paths is excecuted which accounts for all the time that is spent in that function. 

```
[[prod]root@oc-ctl-0 ~]# grep -R 'SELECT aim_endpoint_groups.name' server.log
199
```

In addition to this the long duration of the static path update causes an increase number of StaleData SQLAlchemy exceptions because the chances of a race condition are much higher, which makes scaling up VMs with static paths almost impossible

We have reproduced this issue with a unit test and also have a fix for the scaling issue. I will create a PR.

[1] 
```
Total time: 41.5597 s
/usr/lib/python2.7/site-packages/aim/aim_manager.py
Function: update at line 188
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   188                                               @utils.log
   189                                               @profile
   190                                               def update(self, context, resource, fix_ownership=False,
   191                                                          force_update=False, **update_attr_val):
   192                                                   """Persist updates to AIM resource to the database.
   193
   194                                                   Values of identity attributes of parameter 'resource' are used
   195                                                   to determine the database object to update; no other attributes
   196                                                   from 'resource' are used during the update. Param 'update_attr_val'
   197                                                   specifies the values of the attributes to update.
   198                                                   If the object does not exist in the database, no changes are
   199                                                   made to the database.
   200                                                   """
   201         1          9.0      9.0      0.0          self._validate_resource_class(resource)
   202         1         61.0     61.0      0.0          with context.store.begin(subtransactions=True):
   203         1     282721.0 282721.0      0.7              db_obj = self._query_db_obj(context.store, resource)
   204         1          2.0      2.0      0.0              if db_obj:
[...]

PR #310, opened 2020-10-29
Fix for scaling issue of static paths

timeu:
This is a fix for the scaling issue on in #309.

A new testcase `test_static_path_scale ` was added to test the scaling of static paths. 
In the testcase we add 199 static paths and then in a second step add one new static path, remove one existing and update another one. 

The fix for this issue is to update the static paths in place instead of re-creating them every time even when only one new static path is added.

Here are some numbers:

Runtime of the `test_static_path_scale` test:  

- Fixed version:      2.2s
- Original version: 10.9s 

This is also visible in the number of SELECT statements on the aim_endpoint_groups table:
- Fixed version:         7 
- Original version: 206

### DIFF
--- a/aim/db/models.py
+++ b/aim/db/models.py
@@ -390,13 +390,28 @@ class EndpointGroup(model_base.Base, model_base.HasAimId,
         self.physical_domains = physical_domains
 
         if 'static_paths' in res_attr:
-            static_paths = []
-            for p in (res_attr.pop('static_paths', []) or []):
-                if p.get('path') and p.get('encap'):
-                    static_paths.append(EndpointGroupStaticPath(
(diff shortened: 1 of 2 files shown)


tbachman (changes requested):
This looks great, and a huge thanks for the PR! Only two minor change requests here:
1) Make sure that the line lengths don't exceed 80 characters
2) squash the commits into a single one

Otherwise, this looks good to me , and thanks!

timeu:
@tbachman thx! will do later today

timeu:
@tbachman  squashed commits and made pep8 pass cleanly.

tbachman:
Thanks for the patch!

PR #317, opened 2021-01-19
Fix cert-based authentication

tbachman:
When using certificate-based authentication with APIC, APIC doesn't
require periodic logins to refresh authentication, as authentication
is instead managed by certificates. Password-based authentication uses
a periodic login thread in acitoolkit to maintain valid authentication
session with APIC. However, acitoolkit doesn't spawn a login thread if
certificate-based authentication is used. The AIM/AID code currently
checks to make sure that there is always a login thread present, and
if there isn't, it restarts AIM/AID. This patch fixes AIM/AID to only
check the login thread if username/password-based authentication is
used.

### DIFF
--- a/aim/agent/aid/universes/aci/aci_universe.py
+++ b/aim/agent/aid/universes/aci/aci_universe.py
@@ -339,9 +339,14 @@ class WebSocketContext(object):
         aim_context = aim_ctx.AimContext(store=api.get_store())
         LOG.debug("Monitoring threads login and subscription")
         try:
+            if (self.private_key_file or self.cert_name):
+                threads = ((self.subs_thread, 'subscription_thread'),)
+            else:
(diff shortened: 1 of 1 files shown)


timeu:
FYI:
We ran into this issue when switching to cert based auth some time ago and might have also caused one outage because of a race condition (we had a TAC case for that). 
With the cert based authentication, the aim agent was constantly restarted because [killed](https://github.com/noironetworks/aci-integration-module/blob/0.14.13/aim/agent/aid/universes/aci/aci_universe.py#L348) and then restarted by supervisord

sayalinaval (approved):

tbachman:
Please don't merge yet -- needs another spin - doesn't completely fix.

tbachman:
Should be ready for review now.

irathore (approved):

PR #318, opened 2021-01-19
Avoid mysql 1.0 for py27

tbachman:
The mysql package 1.0 no longer supports python2.7. As a result,
use an older version only for UTs (upstream distros will still
support mysql 1.0 with python2.7).


sayalinaval (approved):

irathore:
+1

PR #320, opened 2021-01-22
Fix ERSPAN reconciliation

tbachman:
The hash tree comparison fails due to integer types being converted
to string values when creating the APIC-side hash tree. These string
values result in different hashes in the hash tree for those resources,
and therefore AID never reaches sync. This patch converts the integer
values used by ERSPAN resources to strings.


tanyatukade (approved):

sayalinaval (approved):

aim/db/migration/alembic_migrations/versions/105b5f35b8fd_erspan_summary_string.py
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) 2020 Cisco Systems
```
tanyatukade, line 1: Minor nit- year should be 2021

tbachman: Good catch - will fix.

tbachman:
Pushed update to address reviewer comments.

PR #321, opened 2021-01-25
Fix Netflow reconciliation

tanyatukade:
The hash tree comparison fails due to integer types being converted to string values when creating the APIC-side hash tree. These string values result in different hashes in the hash tree for those resources, and therefore AID never reaches sync. This patch converts the integer values used by Netflow Rs object to strings.

Signed-off-by: Tanya Tukade tanyatukade.123@gmail.com

### DIFF
--- a/aim/api/resource.py
+++ b/aim/api/resource.py
@@ -1485,17 +1485,17 @@ class VmmRelationToExporterPol(AciResourceBase):
         ('netflow_path', t.string()))
     other_attributes = t.other(
         ('monitored', t.bool),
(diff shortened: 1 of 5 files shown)


tanyatukade:
- The tox pep8 test fails because the db migration script uses previous PR's migration down revision as the revision ID.
- I will wait until the previous PR is merged into master and rerun the test.

tbachman (approved):
Looks good to me. We can rebase this with the new DB HEAD once the ERSPAN patch merges.

tanyatukade:
Rebased with master

PR #323, opened 2021-01-26
Fix QoS reconciliation

tbachman:
The hash tree comparison fails due to integer types being converted
to string values when creating the APIC-side hash tree. These string
values result in different hashes in the hash tree for those resources,
and therefore AID never reaches sync. This patch converts the integer
values used by QoS resources to strings.


tbachman:
This can't be merged into other patches before it go through, but I wanted to put it out so that others can start to review it.

tanyatukade (approved):

PR #326, opened 2021-02-25
Deny action support for contracts

mpaidipa-aci:
A new AIM resource ContractSubjectRsFilter is introduced as a child of
ContractSubject, which replaces bifilter, infilter and outfilter fields

Direction field in the resource will control the in,out and bi types, and
action to store permit/deny


tbachman (changes requested):
Overall, a really good patch! One minor comment on the DB migration. My only other concern with this is that we've tested the upgrade path. I'd like to see the equivalent GBP patch before deciding to merge this. I'd also prefer to see the successful upgrade of a setup with this change and the corresponding GBP change (i.e. hot-patch and run the migrations, then verify operation).

aim/db/migration/alembic_migrations/versions/185efc4806b2_contract_subject_filter.py
```diff
@@ -0,0 +1,159 @@
+# Copyright (c) 2017 Cisco Systems
+# All Rights Reserved.
```
sbagalur, line 2: Fix the copyright year

aim/db/migration/data_migration/contract_subject_filter.py
```diff
@@ -0,0 +1,205 @@
+# Copyright (c) 2018 Cisco Systems
```
sbagalur, line 1: Fix the copyright year

PR #332, opened 2021-07-31
Hierarchical Security Group Changes

sayalinaval:
Changes for adding System Security Group resources and commands in AIM.


irathore (approved):

tbachman:
Please squash commits.

ankur-sao:
Pls don't merge this yet. We are still trying to see if we can have this command without arguments (tenant, name, subject).

ansao-aci:
Pls don't merge this yet. We are still trying to see if we can have this command without arguments (tenant, name, subject).

tbachman:
Thanks for this - but can you close it instead? Sanya used your patch as the starting point for hers, which includes the converter changes. See https://github.com/noironetworks/aci-integration-module/pull/333.

PR #333, opened 2021-09-02
Provide AIM support for Multi Layer Security Group

sanygupt:
This diff contains changes to add new resources to support operator
level security controls that are transparent to tenants.


tbachman:
Looks like you have some formatting issues (see the PEP8 results from TravisCI)?

tbachman:
Also - please change the title: since this is an open source project, we shouldn't reference Cisco-specific bug IDs.

sanygupt:
> Looks like you have some formatting issues (see the PEP8 results from TravisCI)?

Fixed it now,

sanygupt:
> Also - please change the title: since this is an open source project, we shouldn't reference Cisco-specific bug IDs.

Resolved.

aim/aim_manager.py
```diff
@@ -177,6 +180,8 @@ def create(self, context, resource, overwrite=False, fix_ownership=False):
                     context.store.from_attr(old_db_obj, type(resource),
                                             attr_val)
             db_obj = old_db_obj or context.store.make_db_obj(resource)
+            if 'name' in dir(db_obj):
```
tbachman, line 183: Isn't this done already in the resource's constructor object (which is needed in order to create the entry in the DB)?

aim/api/resource.py
```diff
@@ -1751,3 +1751,103 @@ def __init__(self, **kwargs):
...
+        ('direction', t.enum("", "ingress", "egress")),
+        ('ethertype', t.enum("", "undefined", "ipv4", "ipv6")),
+        ('remote_ips', t.list_of_strings),
+        ('ip_protocol', t.string()),
+        ('from_port', t.port),
+        ('to_port', t.port),
+        ('conn_track', t.enum('normal', 'reflexive')),
+        ('icmp_type', t.string()),
+        ('icmp_code', t.string()),
+        ('remote_group_id', t.name),
```
tbachman, line 1831: I'm thinking we should get rid of this for hierarchical security groups - the workflow is for the operator to specify CIDRs that are allowed, not remote groups.

aim/db/migration/alembic_migrations/versions/5b00b6d805e9_system_security_groups.py
```diff
@@ -0,0 +1,111 @@
...
+        sa.Column('direction', sa.String(16)),
+        sa.Column('ip_protocol', sa.String(16)),
+        sa.Column('from_port', sa.String(16)),
+        sa.Column('to_port', sa.String(16)),
+        sa.Column('monitored', sa.Boolean, nullable=False, default=False),
+        sa.Column('epoch', sa.BigInteger(), nullable=False,
+                  server_default='0'),
+        sa.Column('conn_track', sa.String(25), nullable=False,
+                  server_default='reflexive'),
+        sa.Column('remote_group_id', sa.String(64),
```
tbachman, line 85: See above - we probably should remove remote_group_id.

aim/db/models.py
```diff
@@ -1597,3 +1597,97 @@ class SpanSpanlbl(model_base.Base, model_base.HasDisplayName,
...
+                                  lazy='joined')
+    direction = sa.Column(sa.String(16))
+    ethertype = sa.Column(sa.String(16))
+    ip_protocol = sa.Column(sa.String(16))
+    from_port = sa.Column(sa.String(16))
+    to_port = sa.Column(sa.String(16))
+    conn_track = sa.Column(sa.String(25))
+    icmp_code = sa.Column(sa.String(16))
+    icmp_type = sa.Column(sa.String(16))
+    remote_group_id = sa.Column(sa.String(64), nullable=False)
```
tbachman, line 1666: and again here - no remote_group_id.

tbachman (changes requested):
Overall, a really clean patch! Just a few comments.

aim/aim_manager.py
sanygupt: This change was added by @sayalinaval . I think it would be better if she explains this change.

sayalinaval: This can be removed.
And also line 245-246,273-274 should be removed from aim/tests/unit/test_aim_manager.py

sanygupt: Thank you @sayalinaval , will do

tbachman (approved):

PR #334, opened 2021-09-03
Pin oslo.db for python3

tbachman:
With the newer package of oslo.db (11.0.0), running the DB migration
results in the following exception:

    TypeError: 'int' object is not iterable

Pin oslo.db for now, until we can find out why the newer version
breaks it.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -22,6 +22,7 @@ sphinx!=1.2.0,!=1.3b1,<1.3,>=1.1.2
 oslosphinx>=2.5.0 # Apache-2.0
 oslotest>=1.10.0 # Apache-2.0
 oslo.serialization!=2.19.1,>=2.18.0 # Apache-2.0
+oslo.db<=10.0.0 ;python_version!='2.7' # MIT License
 testrepository>=0.0.18
 testscenarios>=0.4
 testtools>=1.4.0


sanygupt (approved):

tbachman:
I debugged this a bit more, and I'm pretty sure that it's a bug in the upstream libraries. The code is trying to do a "SELECT(1)", which is to establish the connection to the DB:
(Pdb) l
 68  	    save_should_close_with_result = connection.should_close_with_result
 69  	    connection.should_close_with_result = False
 70  	    try:
 71  	        # run a SELECT 1.   use a core select() so that
 72  	        # any details like that needed by the backend are handled.
 73  ->	        connection.scalar(select(1))
 
This gets interpreted later as an integer in this for loop:
> /home/vagrant/work/openstack/noiro/aci-integration-module/.tox/pep8/lib/python3.6/site-packages/sqlalchemy/sql/selectable.py(3120)__init__()
(Pdb) l 3110
3105
3106 	        if cols_present:
3107 	            self._raw_columns = []
3108 	            try:
3109 	                for c in columns:

You can see columns gets passed the "1" from the select:
> /home/vagrant/work/openstack/noiro/aci-integration-module/.tox/pep8/lib/python3.6/site-packages/sqlalchemy/util/deprecations.py(139)warned()
(Pdb) kwargs
{'columns': 1, 'whereclause': None, 'from_obj': None, 'distinct': False, 'having': None, 'correlate': True, 'prefixes': None, 'suffixes': None}

We also have the version of SQLAlchemy pinned, so it's possible we need to relax that constraint (i.e. newer version of oslo.db needs to use newer version of SQLAlchemy).

tbachman:
I found that un-pinning the version of SQLAlchemy (goes to version 1.4.23) avoids this issue... but hits a different one, which is the reason we pinned it in the first place. It gets stuck in this statement:
INFO  [alembic.runtime.migration] Running upgrade 2f05b6baf008 -> 61db5ac02ffa, Tables for adding Out of Band BRC
Fixing ownership of no-NAT L3Outs
Rebuilding hash-trees

tbachman:
I figured out what the root issue is here. When rebuilding the hashtrees (done as part of the DB migraiton, which gets run with the PEP8 check), with the newer version of SQLAlchemy, there is an exception thrown:
"'Query' object has no attribute 'with_lockmode'"

We need to remove the use of with_lockmode, since our HA database (usually Galera) doesn't support distributed locking anyway. I verified that removing these lines fixes the hang problem with newer SQLAlchemy:
https://github.com/noironetworks/aci-integration-module/blob/master/aim/aim_store.py#L372-L373

So, while this fix is a temporary workaround, we should provide a newer patch that removes this pin, removes the above lines of code, and removes the SQLAlchemy pin. This change should be thoroughly retested.

tbachman:
Closing, in favor of pull-request #335, which is the correct fix for the issue.

PR #335, opened 2021-09-03
Remove use of with_lockmode

tbachman:
The use of the now-removed with_lockmode parameter is meaningless,
since most deployments use a distributed database (e.g. Galera)
that doesn't support distributed locking.

Commit 462b1ea711c91b06bd5e8c5cfcd3e8c3a1d480e5 pinned SQLAlchemy
in order to avoid the exception that was thrown due to the fact that
the "with_lockmode" parameter was removed from newer versions.

This patch un-pins SQLAlchemy and removes the use of with_lockmode.

### DIFF
--- a/aim/aim_store.py
+++ b/aim/aim_store.py
@@ -369,8 +369,6 @@ class SqlAlchemyStore(AimStore):
             else:
                 args = [getattr(db_obj_type, order_by)]
             query = query.order_by(*args)
-        if lock_update:
-            query = query.with_lockmode('update')
         return query
 
     def query_statuses(self, resources):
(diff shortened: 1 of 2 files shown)


tbachman:
Note that the "lock_update" parameter should be removed from AIM as well, but only after upstream dependencies have removed it's use (e.g. https://opendev.org/x/group-based-policy/src/branch/master/gbpservice/neutron/services/grouppolicy/drivers/cisco/apic/aim_validation.py#L571, which technically is a replacement for AIM, but the prototype is used).

sbagalur (approved):

sanygupt (approved):

PR #338, opened 2021-10-21
Prevent the copy of the empty state when the tenant isn't "warm"

sayalinaval:
Add condition to prevent the copy of the empty state
when the tenant isn't "warm".


tbachman:
I think we can abandon this change - the original issue we were worried about won't happen.

tbachman:
No longer needed.

PR #339, opened 2021-11-04
Restart if tenant manager objects are orphaned

sbagalur:
If there are failures when we 'serve' the tenants in different universes, some orphaned tenant manager objects can exist. To avoid this situation, we restart the process on failure

### DIFF
--- a/aim/agent/aid/universes/aci/aci_universe.py
+++ b/aim/agent/aid/universes/aci/aci_universe.py
@@ -517,10 +517,10 @@ class AciUniverse(base.HashTreeStoredUniverse):
                     serving_tenants[added].start()
         except Exception as e:
             LOG.error(traceback.format_exc())
-            LOG.error('Failed to serve new tenants %s' % tenants)
-            # Rollback served tenants
-            serving_tenants = serving_tenant_copy
-            raise e
+            # There can be orphaned tenant manager objects, If there is a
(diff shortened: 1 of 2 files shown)


aim/agent/aid/universes/aci/aci_universe.py
```diff
@@ -518,9 +518,9 @@ def serve(self, context, tenants):
         except Exception as e:
             LOG.error(traceback.format_exc())
             LOG.error('Failed to serve new tenants %s' % tenants)
-            # Rollback served tenants
-            serving_tenants = serving_tenant_copy
-            raise e
+            # There can be orphaned tenant manager objects, If there is a
+            # failure in serve method. Restart the process
+            utils.perform_harakiri(LOG, "Error in serve. Reset the tenants")
```
tbachman, line 523: Just a nit - we probably shouldn't say "Reset the tenants", because that's not what's happening here. It's a complete restart of AIM. We might want to instead use the message from line 520 above (and get rid of that log message).

tbachman (changes requested):
Just

aim/agent/aid/universes/aci/aci_universe.py
sbagalur: Will do

tbachman (approved):

PR #349, opened 2022-04-03
Make ExternalNetwork contracts top-level resources

tbachman:
The contract resolver (RS) MOs under the external network/EPG of a
Layer3 Outside policy currently aren't top-level AIM resources, and instead
are created or destroyed by adding or removing them to/from lists in the
AIM ExternalNetwork resource. This causes problems when the
ExternalNetwork resource is monitored, and the contract resolver MOs are
configured MOs (i.e. owned and managed by AIM) - if AIM thinks that the
ExternalNetwork has been removed from APIC, it will delete the contract
resolvers from the AIM DB, which causes AID to delete them from ACI.
By promoting the contract resolvers to top level resources, they can
exist in the AIM DB without the need of the parent resource
(ExternalNetwork), which means if the monitored resource goes absent
from the AIM DB, it won't automatically delete the contract resolvers
under it.


tbachman:
I'm going to update the PR to add UTs for the new AIM resources. I could also do the same refactoring for contracts under EPG, or we can do that in a separate patch (open to thoughts on this one).

irathore (approved):

tbachman:
Pushed latest updates to PR - ready for review.

sayalinaval (approved):

aim/db/migration/alembic_migrations/versions/7b8a71bee019_ext_epg_contracts.py
```diff
@@ -0,0 +1,70 @@
+# Copyright (c) 2020 Cisco Systems
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
```
sayalinaval, line 4: Shouldn't the file name be '7b8a71bee019_ext_net_contracts.py'?

tbachman: It doesn't have to, but I should change it just so it's consistent with the naming used in the patch. Thanks!

tbachman:
This has now been functionally tested, and ready for review.

irathore (approved):

sayalinaval (approved):

PR #359, opened 2022-08-31
change mask of /run/aid/events to 0777

irathore:


tbachman:
Do we still need this? If not, can we close it?

PR #379, opened 2023-02-23
Removing persistance of hashtrees in DB

sbagalur:
This commit removes persisting the hashtrees in DB. Instead the hashtrees are saved in memory. This helps reduce the space issues when there are is a large set of config on the device. We still save empty hashtrees in the DB. This is done to make the impact minimal and not to affect the current workflow of AIM.


christides11 (dismissed):

tbachman (changes requested):
Overall, nicely done! One general comment - it would be helpful to have a comment message that explains what's being done at a high level (e.g. introduction of the in-memory hash tree, but still using the DB for the hash tree roots), along with an explanation for why we're moving away from persisting hash trees (e.g. hash tree BLOBs in the DB were getting too big, and interested in reducing DB load).

PR #380, opened 2023-05-08
L3Outside Multi External Network Support

christides11:
Support for alternative naming scheme for epgs.

### DIFF
--- a/aim/aim_lib/nat_strategy.py
+++ b/aim/aim_lib/nat_strategy.py
@@ -106,7 +106,8 @@ class NatStrategy(object):
 
     @abc.abstractmethod
     def create_l3outside(self, ctx, l3outside,
-                         vmm_domains=None, phys_domains=None):
+                         vmm_domains=None, phys_domains=None,
+                         epg_name=None):
         """Create L3Outside object if needed.
 
         :param ctx: AIM context
@@ -115,7 +116,7 @@ class NatStrategy(object):
         """
 
     @abc.abstractmethod
-    def delete_l3outside(self, ctx, l3outside):
+    def delete_l3outside(self, ctx, l3outside, epg_name=None):
         """Delete L3Outside object.
 
         :param ctx: AIM context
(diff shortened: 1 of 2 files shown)


test-requirements.txt
```diff
@@ -29,3 +29,4 @@ testtools>=1.4.0
 mock
 tabulate>=0.7.5
 voluptuous!=0.13.0
+SQLAlchemy==1.4.47
```
tbachman, line 32: Since we use AIM in different distros (Ubuntu 18.04, 20.04, RHEL7, RHEL8), I don't think we should specify the constraint this way. Instead, we may want to limit it to be below a certain version. That may take experimentation. We also should probably do that as a separate patch.

aim/aim_lib/nat_strategy.py
```diff
@@ -624,23 +633,25 @@ def _select_domains(self, objs, vmm_domains=None, phys_domains=None):
                 if phys_domains is not None:
                     obj.physical_domains = phys_domains
 
-    def _create_nat_epg(self, ctx, l3out, vmm_domains=None, phys_domains=None):
-        objs = self._get_nat_objects(ctx, l3out)
+    def _create_nat_epg(self, ctx, l3out, vmm_domains=None, phys_domains=None,
+                        epg_name=None):
+        objs = self._get_nat_objects(ctx, l3out, epg_name)
```
tbachman, line 638: This is a nit, but you probably should pass this as a named kwarg: epg_name=epg_name. That would guard against any changes in the future that might introduce more positional arguments.

tbachman (changes requested):
Pretty much LGTM, but a few nits. We probably want to break out the test-requirements.txt into a separate patch, and then just a NIT in one of the calls.

sayalinaval (approved):

test-requirements.txt
```diff
@@ -28,4 +28,4 @@ testscenarios>=0.4
 testtools>=1.4.0
 mock
 tabulate>=0.7.5
-voluptuous!=0.13.0
+voluptuous!=0.13.0
```
tbachman, line 31: Is there a change here? If not, you should be able to remove any changes to this file.

tbachman (approved):
LGTM!

sayalinaval (approved):

tbachman (approved):

PR #388, opened 2023-10-04
Add GARP configuration for gen1 hardware

tbachman:
First generation switches are unable to use GARPs for EP movement detection in hardware when the EP "move" is on the same port..As a result, additional configuraiton is needed in the BD to punt all GARPs to the switch CPU for processing. This patch adds an AIM parameter to control how this behavior is configured in the fabric, with the default configuration to disable it (i.e. default is to not support use of GARPs for EP movement detection on gen1 hardware).


sayalinaval (approved):

tbachman:
Hold off on merging this - I want to create a data migration DB migration, which changes the values of existing BDSs.

tbachman:
Just updated the test with the data migration (and UT that tests it). Ready for reviews.

christides11 (approved):

sayalinaval (approved):

tbachman:
I just realized - the data migration doesn't consider the configuration variable. Let me fix that and re-spin.

tbachman:
Just pushed latest changes.

sbagalur (approved):

PR #389, opened 2023-10-05
Add scope attribute for subnet create

sayalinaval:


tbachman (changes requested):
I think to keep backwards compatibility, you want to use a named arg with a default value - like scope='public'. That way old code works with the new API, which uses the current default, but new code can set the value differently by passing "scope='public,shared'". Also, some UTs failed, but it's unclear if it's related (seems to be an ordering issue?).

aim/aim_lib/nat_strategy.py
```diff
@@ -135,7 +135,7 @@ def get_l3outside_resources(self, ctx, l3outside, epg_name=None):
         """
 
     @abc.abstractmethod
-    def create_subnet(self, ctx, l3outside, gw_ip_mask):
+    def create_subnet(self, ctx, l3outside, gw_ip_mask, scope):
```
tbachman, line 138: The abstract base class methods. should have the same signature as their concrete class methods.

tbachman (approved):

PR #398, opened 2023-11-14
Normalization of Remote IP feature support

sbagalur:
Change-Id: Icf3d76f9945b5c18b119ddf7b4d5cc3bf9a853fe

This patch adds support for Normalization of Remote IP feature support


aim/api/resource.py
```diff
@@ -1223,6 +1223,7 @@ class SecurityGroupRule(AciResourceBase):
         ('icmp_type', t.string()),
         ('icmp_code', t.string()),
         ('remote_group_id', t.name),
+        ('tDn', t.string(256)),
```
tbachman, line 1226: Would an empty string make sense as a default value for this field?

tbachman (approved):
Looks good to me - the only question/comment I have is whether it makes sense to have a default value of an empty string for the tDn field.

aim/api/resource.py
sbagalur: It is only set by GBP, and Sayali sets it to empty string I believe, when it is not updated. I can add it to the init as well.

christides11 (approved):

tbachman (approved):

PR #402, opened 2024-01-05
POST delete instead of REST DELETE command

sayalinaval:


tbachman:
Please make sure to squash the commits once you're done.

aim/agent/aid/universes/aci/tenant.py
```diff
@@ -618,22 +624,25 @@ def _push_aim_resources(self):
...
                                     self.ac_context.aci_session.DELETE(
                                         '/mo/%s.json' % attr.pop('dn'))
+                                LOG.debug("DELETING from APIC: %s" % to_delete)
+                                self._post_with_transaction(to_delete,
+                                                            deleted=True)
                                 LOG.debug("UPDATING in APIC: %s" % to_update)
                                 # Update object ownership
                                 self._post_with_transaction(to_update,
                                                             modified=True)
-                                if to_update:
```
tbachman, line 635: I'm not sure I follow the logic here to make this conditional always happen. Before, it would only happen if to_update was true?

tbachman (commented):
Other than the question about the creation_succeeded call change, LGTM (marking it with no changes requested, but would like to get an answer to the question before merging).

aim/agent/aid/universes/aci/tenant.py
sayalinaval: Before it would happen for create and update. It would be good for testing purposes to use it for delete as well.

tbachman (approved):

PR #407, opened 2024-02-01
Make StaticPath top level resource

sonianuj287:


tbachman:
We'll use https://github.com/noironetworks/aci-integration-module/pull/409, as it's further along.

PR #409, opened 2024-02-15
Make static path top level resource

sayalinaval:


irathore (dismissed):

sbagalur (dismissed):

aim/db/migration/alembic_migrations/versions/e322787e56fd_modify_static_paths_table.py
```diff
@@ -0,0 +1,70 @@
...
+from aim.db.migration.data_migration import static_paths_v2
+
+revision = 'e322787e56fd'
+down_revision = '79deb77b4719'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
```
tbachman, line 36: We may want to wrap all of this in a context manager. That way, either the entire DB operation (new table creation, data migration, table rename) happens atomically or gets rolled back. Otherwise, end-users would be required to manually fix the DB in order to run the DB migration again.

tbachman (changes requested):
Some minor nits.

christides11 (approved):

aim/db/migration/alembic_migrations/versions/e322787e56fd_modify_static_paths_table.py
sayalinaval: done

PR #413, opened 2024-04-17
Command to print hashtree node count

christides11:


aim/tools/cli/debug/hashtree.py
```diff
@@ -65,6 +65,35 @@ def dump(ctx, tenant, flavor):
             click.echo('%s not found for tenant %s' % (search.__name__, t))
 
 
+@hashtree.command(name='dump-amt')
```
tbachman, line 68: This is a real nit-pick, but can we change the command to something a bit more descriptive? Maybe "count-tree-nodes"?

tbachman (commented):
Small nit....

tbachman (approved):

PR #433, opened 2024-11-28
Rebase SqlAlchemy_2.0 Support

nisar123456:


aim/agent/aid/universes/aim_universe.py
```diff
@@ -115,11 +116,9 @@ def get_optimized_state(self, context, other_state,
...
         # Only delete if state is still empty. Never remove a tenant if there
         # are leftovers.
-        with context.store.begin(subtransactions=True):
-            # There could still be logs, but they will re-create the
-            # tenants in the next iteration.
-            self.tree_manager.delete_by_root_rn(context, key, if_empty=True)
-        super(AimDbUniverse, self).cleanup_state(context, key)
+        # There could still be logs, but they will re-create the
+        # tenants in the next iteration.
+        self.tree_manager.delete_by_root_rn(context, key, if_empty=True)
```
tbachman, line 121: The previous patch called the parent class's cleanup_state method. Isn't that still needed here?

aim/agent/aid/service.py
```diff
@@ -235,51 +235,51 @@ def _send_heartbeat(self, aim_ctx):
...
-                agents = [x for x in agents if
-                          self._major_vercompare(x.version, max_version) == 0]
-            else:
-                agents = [self.agent]
-            result = self._tenant_assignation_algorithm(aim_ctx, agents)
-            # Store result in DB
-            self.agent.hash_trees = result
-            self.manager.create(aim_ctx, self.agent, overwrite=True)
-            return result
+        # Refresh this agent
```
tbachman: I think this is a problem - the agent update is not scoped by a transaction any more, and therefore can't be atomic. What was the reason to remove the transaction here?

aim/agent/aid/universes/aim_universe.py
```diff
@@ -228,33 +227,32 @@ def _push_resource(self, context, resource, method, monitored):
...
-                            **ext(resource, "other"))
-                        if obj:
-                            # Declare victory for the update
-                            self.creation_succeeded(resource)
-                    else:
-                        self.manager.create(
-                            context, resource, overwrite=True,
-                            fix_ownership=monitored)
-                        # Declare victory for the created object
+                if isinstance(resource, aim_resource.AciRoot):
```
tbachman: Same concern here - the update is no longer atomic. What was the reason to remove the transaction?

aim/agent/aid/universes/aim_universe.py
```diff
@@ -264,10 +262,9 @@ def _push_resource(self, context, resource, method, monitored):
                     return
                 if monitored:
                     # Only delete a resource if monitored
-                    with context.store.begin(subtransactions=True):
-                        existing = self.manager.get(context, resource)
-                        if existing and existing.monitored:
-                            self.manager.delete(context, resource)
+                    existing = self.manager.get(context, resource)
```
tbachman: And here - the get/delete code could be concurrent, and therefore needs the protection of a transaction.

aim/aim_lib/nat_strategy.py
```diff
@@ -437,97 +437,93 @@ def _create_l3out(self, ctx, l3out, vmm_domains=None, phys_domains=None,
...
-                    self.mgr.create(ctx, ext_vrf)
-                l3out_db = copy.copy(l3out)
-                l3out_db.vrf_name = ext_vrf.name
-                l3out_db = self.mgr.create(ctx, l3out_db)
-            self._create_nat_epg(ctx, l3out_db,
-                                 vmm_domains=vmm_domains,
-                                 phys_domains=phys_domains,
-                                 epg_name=epg_name)
-            return l3out_db
+        tenant = resource.Tenant(name=l3out.tenant_name)
```
tbachman: Same concern here.

tbachman (changes requested):
I didn't look through the entire patch - there are a lot of cases where code that was previously safe under concurrency would no longer be protected. I'd like to get a better understanding of why the concurrency there was removed before going further in this patch.

aim/agent/aid/universes/aim_universe.py
nisar123456: yeah, parent class's cleanup_state method required. i will add it back

aim/agent/aid/service.py
nisar123456: Since _calculate_tenants() is always invoked with an active aim_ctx whose Session is already in a transactional state, the explicit aim_ctx.store.begin(subtransactions=True) wrapper was removed.
The self.manager.create() call still executes inside the caller-managed outer transaction, so the agent update remains atomic.

tbachman (approved):

PR #451, opened 2025-09-10
osp18 agent id name changes

sbagalur:
In OSP18, since pods names are ephemeral, we want to modify the naming of AID agents, so that the naming doesnt change if redeployed. This would make sure theres no thrashing of tenants assignments.


tbachman (changes requested):
One minor comment, otherwise good - thanks for the patch!

tbachman (approved):

PR #458, opened 2025-11-05
Enhancing the ssl options to be Bool/Str path

sbagalur:
Enhancing ssl options to support bool and string path.

It now supports True, False or a custom CA path string.


tbachman (approved):
Approved, but we should re-check once the apicapi patch has been merged to verify that it passes the gate.

